PR #3, opened 2020-07-20
Workspaces related stuffs

ashwin1111:


apps/workspaces/models.py
```diff
@@ -0,0 +1,44 @@
+"""
+Workspace Models
+"""
+from django.db import models
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+
+class Workspace(models.Model):
```
Shwetabhk, line 10: Please rename the tables. We did not do in Quickbooks, so we have weird table names in QBO
Check this for example
https://github.com/fylein/fyle-netsuite-api/blob/master/apps/workspaces/models.py

apps/workspaces/models.py
```diff
@@ -0,0 +1,44 @@
...
+    id = models.AutoField(primary_key=True)
+    si_user_id = models.TextField(help_text='Stores Sage Intacct user id')
+    si_company_id = models.TextField(help_text='Stores Sage Intacct company id')
+    si_user_password = models.TextField(help_text='Stores Sage Intacct user password')
+    workspace = models.OneToOneField(Workspace, on_delete=models.PROTECT, help_text='Reference to Workspace model')
+    created_at = models.DateTimeField(auto_now_add=True, help_text='Created at datetime')
+    updated_at = models.DateTimeField(auto_now=True, help_text='Updated at datetime')
+
+
+class FyleCredential(models.Model):
```
Shwetabhk, line 36: Same

apps/workspaces/models.py
```diff
@@ -0,0 +1,44 @@
...
+    id = models.AutoField(primary_key=True, help_text='Unique Id to identify a workspace')
+    name = models.CharField(max_length=255, help_text='Name of the workspace')
+    user = models.ManyToManyField(User, help_text='Reference to users table')
+    fyle_org_id = models.CharField(max_length=255, help_text='org id', unique=True)
+    last_synced_at = models.DateTimeField(help_text='Datetime when expenses were pulled last', null=True)
+    created_at = models.DateTimeField(auto_now_add=True, help_text='Created at datetime')
+    updated_at = models.DateTimeField(auto_now=True, help_text='Updated at datetime')
+
+
+class SageIntacctCredential(models.Model):
```
Shwetabhk, line 23: Same

apps/workspaces/views.py
```diff
@@ -0,0 +1,295 @@
...
+                    user_password=si_user_password
+                )
+
+                sage_intacct_credentials = SageIntacctCredential.objects.create(
+                    si_user_id=si_user_id,
+                    si_company_id=si_company_id,
+                    si_user_password=si_user_password,
+                    workspace=workspace
+                )
+            else:
```
Shwetabhk, line 229: Why do we need an else case here?
Also we're not checking in the else case whether the SDK is working correctly with the new credentials or not

apps/workspaces/views.py
```diff
@@ -0,0 +1,295 @@
...
+        except sage_intacct_exc.InternalServerError:
+            return Response(
+                {
+                    'message': 'Wrong/Expired Authorization code'
+                },
+                status=status.HTTP_401_UNAUTHORIZED
+            )
+
+    def delete(self, request, **kwargs):
+        """Delete credentials"""
```
Shwetabhk, line 268:
```
"""
Delete credentials
"""
```

apps/workspaces/views.py
ashwin1111: we need an else case here, when the company has either deleted the user / changed the password, ideally they should be able to change credentials. Yes, will have to add condition to check whether the creds is right / wrong.

Shwetabhk: just call the SDK after if / else in this view, if the creds are correct it will return 200 else 400

PR #9, opened 2020-07-28
General settings and few other minor api fixes

ashwin1111:


apps/fyle/views.py
```diff
@@ -37,7 +37,7 @@ def get_queryset(self):
         elif state == 'READY':
             return ExpenseGroup.objects.filter(
                 workspace_id=self.kwargs['workspace_id'],
-                vendorbill__isnull=True,
+                bill__isnull=True,
```
Shwetabhk, line 40: instead of adding both bill__isnull and expensereport__isnull
can simply use `tasklog__isnull=True`
I think this should work correctly for all cases

apps/tasks/models.py
```diff
@@ -22,12 +22,13 @@ class TaskLog(models.Model):
...
-        help_text='Reference to VendorBill', null=True)
+    bill = models.ForeignKey(Bill, on_delete=models.PROTECT, \
+        help_text='Reference to Bill', null=True)
     expense_report = models.ForeignKey(ExpenseReport, on_delete=models.PROTECT, \
         help_text='Reference to ExpenseReport', null=True)
     status = models.CharField(max_length=255, help_text='Task Status')
     detail = JSONField(help_text='Task Response', null=True, default=get_default)
+    sage_intacct_errors = JSONField(help_text='Sage Intacct Errors', null=True)
     created_at = models.DateTimeField(auto_now_add=True, help_text='Created at datetime')
     updated_at = models.DateTimeField(auto_now=True, help_text='Updated at datetime')
```
Shwetabhk, line 33:
```python
class Meta:
    db_table = 'task_logs' # instead of task_log, all the table names should be plurals
```

PR #16, opened 2020-08-04
Content and few bug fix post deployment

ashwin1111:


apps/sage_intacct/models.py
```diff
@@ -166,7 +166,11 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
         """
         expenses = expense_group.expenses.all()
         bill = Bill.objects.get(expense_group=expense_group)
-        general_mappings = GeneralMapping.objects.get(workspace_id=expense_group.workspace_id)
+
+        try:
```
Shwetabhk, line 170: can use this format
```python
p = GeneralMapping.objects.filter(workspace_id=expense_group.workspace_id).first()
```

This will return none in p if it does not exist

apps/sage_intacct/models.py
```diff
@@ -280,7 +284,11 @@ def create_expense_report_lineitems(expense_group: ExpenseGroup):
         """
         expenses = expense_group.expenses.all()
         expense_report = ExpenseReport.objects.get(expense_group=expense_group)
-        general_mappings = GeneralMapping.objects.get(workspace_id=expense_group.workspace_id)
+
+        try:
```
Shwetabhk, line 288: can use this format
```python
p = GeneralMapping.objects.filter(workspace_id=expense_group.workspace_id).first()
```

This will return none in p if it does not exist

PR #21, opened 2020-10-05
Support for CCC Expenses

ashwin1111:


apps/fyle/views.py
```diff
@@ -48,10 +48,16 @@ def post(self, request, *args, **kwargs):
         """
         state = request.data.get('state', ['PAYMENT_PROCESSING'])
         task_log = TaskLog.objects.get(pk=request.data.get('task_log_id'))
+        general_settings = WorkspaceGeneralSettings.objects.get(workspace_id=kwargs['workspace_id'])
+
+        fund_source = ['PERSONAL']
+        if general_settings.corporate_credit_card_expenses_object is not None:
```
Shwetabhk, line 54: `is not None:` is not needed
just `if general_settings.corporate_credit_card_expenses_object:`
will do

Shwetabhk (approved):

PR #31, opened 2021-01-29
Sage Intacct Project Sync and Auto Map

ashwin1111:


apps/fyle/views.py
```diff
@@ -226,13 +226,12 @@ def post(self, request, *args, **kwargs):
         Get categories from Fyle
```
Shwetabhk, line 226: Get Projects from Fyle

Shwetabhk (approved):

PR #48, opened 2021-03-08
Make use of auto create flag and removing employee_code filter

NileshPant1999:


apps/workspaces/utils.py
```diff
@@ -52,7 +54,7 @@ def create_or_update_general_settings(general_settings_payload: Dict, workspace_
         workspace_id=workspace_id
     )
 
-    if general_settings_payload['auto_map_employees']:
-       schedule_auto_map_employees(general_settings_payload['auto_map_employees'], workspace_id)
+    
+    schedule_auto_map_employees(general_settings_payload['auto_map_employees'], workspace_id)
```
ashwin1111, line 58: can you add ccc related fn as well here

apps/mappings/tasks.py
```diff
@@ -401,7 +401,9 @@ def auto_create_category_mappings(workspace_id):
                     workspace_id=workspace_id,
                 )
                 mapping.source.auto_mapped = True
-                mapping.source.save(update_fields=['auto_mapped'])
+                mapping.source.auto_created = True
+                mapping.source.save(update_fields=['auto_mapped', 'auto_created'])
+
```
ashwin1111, line 406: can you add to project as well

ashwin1111 (approved):

PR #77, opened 2021-05-18
Support for CCC Expenses in Expense Reports

ashwin1111:


apps/sage_intacct/views.py
```diff
@@ -726,3 +726,15 @@ def post(self, request, *args, **kwargs):
             )
 
 
+class ExpenseCustomFieldsView(generics.ListCreateAPIView):
```
Shwetabhk, line 729: can we rename this to ExpenseFields view

Shwetabhk (approved):

Shwetabhk (approved):

PR #78, opened 2021-05-21
support for user defined dimension in sage intacct

NileshPant1999:
Tests Done
1. Tested for both case when mapping single user dimension as well as in case of multiple one.
2. As we have updated the way we export expense_report there are some additional changes that i have to do such as in case of uploading the attachment it was throwing key error as now the response is different then that of earlier one i have fixed this also.


apps/sage_intacct/tasks.py
```diff
@@ -436,10 +436,10 @@ def create_expense_report(expense_group: ExpenseGroup, task_log_id):
                                                                                  expense_report_lineitems_objects)
 
             created_attachment_id = load_attachments(sage_intacct_connection, \
-                                                     created_expense_report['key'], expense_group)
+                                                     created_expense_report['data']['eexpenses']['RECORDNO'], expense_group)
             if created_attachment_id:
```
NileshPant1999, line 440: We are getting different response in via new api that is why changes key here

apps/sage_intacct/views.py
```diff
@@ -725,4 +725,15 @@ def post(self, request, *args, **kwargs):
                 status=status.HTTP_400_BAD_REQUEST
             )
 
+class UserDefinedDimensionsView(generics.ListCreateAPIView):
+    """
+    Expense Custom view
```
NileshPant1999, line 730: Dimensions View

apps/sage_intacct/models.py
```diff
@@ -160,6 +161,44 @@ def get_expense_purpose(workspace_id: int, lineitem: Expense, category: str) ->
         lineitem.employee_email, category, spent_at, lineitem.claim_number, expense_purpose, expense_link)
 
 
+def get_user_defined_dimension_object(expense_group: ExpenseGroup, lineitem: Expense):
+    mapping_settings = MappingSetting.objects.filter(workspace_id=expense_group.workspace_id).all()
+
+    user_dimensions = []
+    default_expense_attributes = ['CATEGORY', 'EMPLOYEE']
+    default_destination_attributes = ['DEPARTMENT', 'LOCATION', 'CLASS', 'PROJECT']
```
ashwin1111, line 169: Do we have CLASS in intacct? Please add all possible attributes. Few examples: 
 EXPENSE_TYPE
 VENDOR
 CCC_ACCOUNT
 CHARGE_CARD_NUMBER
 ITEM
 DEPARTMENT
 PROJECT
 EMPLOYEE
 LOCATION
 ACCOUNT

apps/sage_intacct/models.py
```diff
@@ -160,6 +161,44 @@ def get_expense_purpose(workspace_id: int, lineitem: Expense, category: str) ->
...
+                source_value = lineitem.custom_properties.get(attribute.display_name, None)
+
+            mapping: Mapping = Mapping.objects.filter(
+                source_type=setting.source_field,
+                destination_type=setting.destination_field,
+                source__value=source_value,
+                workspace_id=expense_group.workspace_id
+            ).first()
+            if mapping:
+                dimension_name = 'GLDIM' + mapping.destination.attribute_type
```
ashwin1111, line 192: `'GLDIM{}'.format(mapping.destination.attribute_type)`

apps/sage_intacct/urls.py
```diff
@@ -41,5 +42,6 @@
     path('reimbursements/', ReimbursementView.as_view()),
     path('fyle_reimbursements/', FyleReimbursementsView.as_view()),
     path('sync_dimensions/', SyncSageIntacctDimensionView.as_view()),
-    path('refresh_dimensions/', RefreshSageIntacctDimensionView.as_view())
+    path('refresh_dimensions/', RefreshSageIntacctDimensionView.as_view()),
+    path('user_defined_dimensions/', UserDefinedDimensionsView.as_view())
```
ashwin1111, line 46: not needed, please use this https://github.com/fylein/fyle-intacct-api/pull/75/files#diff-107dc7ac18a460b5b926f7e5b60cefd40ea112b07452c3cdc79d2614e83760c4R729. Send only attribute_type from client.

apps/sage_intacct/utils.py
```diff
@@ -271,6 +271,36 @@ def sync_employees(self):
...
+        """
+        Get User Defined Dimensions
+        """
+        user_defined_dimensions = []
+
+        dimensions = self.connection.dimensions.get()['dimension']
+
+        for dimension in dimensions:
+            if dimension['userDefinedDimension'] == 'true':
+                user_defined_dimensions.append(dimension['objectName'])
```
ashwin1111, line 285: why is user_defined_dimensions needed? Can we bring looping logic written below, inside this if?

apps/sage_intacct/utils.py
```diff
@@ -541,8 +580,8 @@ def __construct_expense_report(self, expense_report: ExpenseReport, \
             'state': 'Submitted',
             'description': expense_report.memo,
             'currency': expense_report.currency,
-            'expenses': {
-                'expense': expsense_payload
+            'EEXPENSESITEMS': {
```
ashwin1111, line 583: keep it as lower if the api accepts, just to maintain consistency

apps/sage_intacct/views.py
```diff
@@ -725,4 +725,15 @@ def post(self, request, *args, **kwargs):
                 status=status.HTTP_400_BAD_REQUEST
             )
 
+class UserDefinedDimensionsView(generics.ListCreateAPIView):
```
ashwin1111, line 728: https://github.com/fylein/fyle-intacct-api/pull/75/files#diff-107dc7ac18a460b5b926f7e5b60cefd40ea112b07452c3cdc79d2614e83760c4R729

apps/sage_intacct/urls.py
NileshPant1999: got it will do this

apps/sage_intacct/utils.py
NileshPant1999: working with smallcase also with make it small

Sravanksk (approved):

Shwetabhk (approved):

PR #83, opened 2021-06-03
Export against employee default location if location is not selected in General Mappings

Sravanksk:


apps/sage_intacct/models.py
```diff
@@ -467,6 +493,12 @@ def create_expense_report_lineitems(expense_group: ExpenseGroup):
                 workspace_id=expense_group.workspace_id
             ).first()
 
+            if general_mappings.use_intacct_employee_locations:
```
ashwin1111, line 496: can we do this at else of this if.
`location_id = get_location_id_or_none(expense_group, lineitem, general_mappings)`

apps/sage_intacct/models.py
```diff
@@ -467,6 +493,12 @@ def create_expense_report_lineitems(expense_group: ExpenseGroup):
...
+                default_employee_location_id = get_intacct_employee_object('location_id', expense_group)
+
+            if general_mappings.use_intacct_employee_departments:
```
ashwin1111, line 499: can we do this at else of this if.
`department_id = get_department_id_or_none(expense_group, lineitem, general_mappings)`

ashwin1111 (approved):

PR #84, opened 2021-06-10
Fix expense reports payload to enable payment sync

ashwin1111:


apps/sage_intacct/tasks.py
```diff
@@ -460,7 +462,10 @@ def create_expense_report(expense_group: ExpenseGroup, task_log_id):
                         expense_group.id, expense_group.workspace_id, {'error': error}
                     )
 
-            task_log.detail = created_expense_report
+            details = {
```
ashwin1111, line 465: this is the actual change, keeping them in the old way

Sravanksk (approved):

Shwetabhk (approved):

PR #89, opened 2021-06-28
store cluster domain in workspace_table

NileshPant1999:


apps/workspaces/views.py
```diff
@@ -54,8 +55,10 @@ def post(self, request):
             workspace.user.add(User.objects.get(user_id=request.user))
             cache.delete(str(workspace.id))
         else:
-            workspace = Workspace.objects.create(name=org_name, fyle_org_id=org_id)
+            fyle_connector = FyleConnector(auth_tokens.refresh_token)
+            cluster_domain = fyle_connector.get_cluster_domain()['cluster_domain']
```
ashwin1111, line 59: Please check app as well, we have a /cluster_domain api which returns it. You'll have to take care of all that. Basically only 1 call per workspace at time of creation should go to fyle. Rest should read from table.

ashwin1111 (changes requested):

apps/workspaces/views.py
NileshPant1999: okay will take care of this.

ashwin1111 (approved):

ashwin1111:
TODO: Have scripts ready before deploy

Sravanksk (approved):

apps/sage_intacct/utils.py
```diff
@@ -485,13 +485,12 @@ def get_expense_link(self, lineitem) -> str:
         :param expense: Expense Lineitem
         :return: Expense link
         """
-        fyle_credentials = FyleCredential.objects.get(workspace_id=self.workspace_id)
-        fyle_connector = FyleConnector(fyle_credentials.refresh_token, self.workspace_id)
-        org_id = Workspace.objects.get(id=self.workspace_id).fyle_org_id
+        workspace = Workspace.objects.get(id=self.workspace_id)
+        org_id = workspace.fyle_org_id
+        cluster_domain = workspace.cluster_domain
```
ashwin1111, line 490: sir, this approach would break for existing clients

ashwin1111 (changes requested):

apps/sage_intacct/utils.py
Shwetabhk:
```python
if not cluster_domain:
   fyle_credentials = FyleCredential.objects.get(workspace_id=self.workspace_id)
   fyle_connector = FyleConnector(fyle_credentials.refresh_token, self.workspace_id)
   cluster_domain = fyle_connector.get_cluster_domain()
   workspace.cluster_domain = cluster_domain
   workspace.save()
```

PR #99, opened 2021-07-19
adding expense link back to memo

NileshPant1999:


apps/sage_intacct/models.py
```diff
@@ -147,12 +147,20 @@ def get_transaction_date(expense_group: ExpenseGroup) -> str:
     return datetime.now().strftime('%Y-%m-%dT%H:%M:%S')
 
 
-def get_expense_purpose(lineitem: Expense, category: str) -> str:
+def get_expense_purpose(workspace_id, lineitem: Expense, category: str) -> str:
+
```
ashwin1111, line 151: rm

apps/sage_intacct/models.py
```diff
@@ -147,12 +147,20 @@ def get_transaction_date(expense_group: ExpenseGroup) -> str:
...
+    workspace = Workspace.objects.get(id=workspace_id)
+    org_id = workspace.fyle_org_id
+    cluster_domain = workspace.cluster_domain
```
ashwin1111, line 154: this would have conflict with the other PR, don't use cluster domain in workspace table here. This should go out tomo, so make this function however we do before (call fyle api and get the cluster domain)

ashwin1111 (changes requested):

apps/sage_intacct/models.py
ashwin1111: Discussed offline, handle not null case @NileshPant1999

apps/sage_intacct/models.py
```diff
@@ -147,12 +147,25 @@ def get_transaction_date(expense_group: ExpenseGroup) -> str:
...
+def get_expense_purpose(workspace_id, lineitem: Expense, category: str) -> str:
+    workspace = Workspace.objects.get(id=workspace_id)
+    org_id = workspace.fyle_org_id
+
+    if workspace.cluster_domain:
+        cluster_domain = workspace.cluster_domain
+    else:
+        fyle_credentials = FyleCredential.objects.get(workspace_id=workspace_id)
+        fyle_connector = FyleConnector(fyle_credentials.refresh_token, workspace_id)
+        cluster_domain = fyle_connector.get_cluster_domain()['cluster_domain']
```
ashwin1111, line 159: Please take care of updating value in the workspace table so that we don't end up at else every time.

ashwin1111 (approved):

PR #109, opened 2021-09-15
adding customer support

labhvam5:


apps/sage_intacct/utils.py
```diff
@@ -1,6 +1,7 @@
 import logging
 from typing import List, Dict
 from datetime import datetime
+from django.db.models import fields
```
ashwin1111, line 4: why is this needed?

apps/sage_intacct/utils.py
```diff
@@ -371,6 +372,27 @@ def sync_classes(self):
...
 
+    def sync_customers(self):
+        """
+        Get Customers
+        """
+        customers = self.connection.customers.get_all()
+
+        customer_attributes = []
+
+        for _customer in customers:
```
ashwin1111, line 383: `_class` is written that way because of the reserved keyword :P
Can use just `customer` here
```
for customer in customers:```

File: apps/sage_intacct/models.py [hunk 3]
```diff
@@ -107,9 +107,36 @@ def get_location_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
     return location_id
 
 
-def get_customer_id_or_none(expense_group: ExpenseGroup, project_id: str):
+def get_customer_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, general_mappings: GeneralMapping, project_id: str):
     customer_id = None
+    if not project_id:
+        if general_mappings and general_mappings.default_customer_id:
```
ashwin1111, line 113: will check and get back if this is needed

apps/sage_intacct/models.py
```diff
@@ -107,9 +107,36 @@ def get_location_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
 
+        customer_setting: MappingSetting = MappingSetting.objects.filter(
+            workspace_id=expense_group.workspace_id,
+            destination_field='CUSTOMER'
+        ).first()    
+
+        if customer_setting:
+            if customer_setting.source_field == 'PROJECT':
+                source_value = lineitem.project
+            elif customer_setting.source_field == 'COST CENTER':
```
ashwin1111, line 124: this should be `COST_CENTER`

apps/sage_intacct/models.py
```diff
@@ -107,9 +107,36 @@ def get_location_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
+            mapping: Mapping = Mapping.objects.filter(
+                source_type=customer_setting.source_field,
+                destination_type='CUSTOMER',
+                source__value=source_value,
+                workspace_id=expense_group.workspace_id
+            ).first()
+
+            if mapping:
+                customer_id = mapping.destination.destination_id    
+
```
ashwin1111, line 139: At the end of if loop,
`return customer_id`

apps/sage_intacct/models.py
```diff
@@ -107,9 +107,36 @@ def get_location_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
+                source_type=customer_setting.source_field,
+                destination_type='CUSTOMER',
+                source__value=source_value,
+                workspace_id=expense_group.workspace_id
+            ).first()
+
+            if mapping:
+                customer_id = mapping.destination.destination_id    
+            
     if project_id:
```
ashwin1111, line 140: `else:`

apps/sage_intacct/models.py
```diff
@@ -107,9 +107,36 @@ def get_location_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
     return location_id
 
 
-def get_customer_id_or_none(expense_group: ExpenseGroup, project_id: str):
+def get_customer_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, general_mappings: GeneralMapping, project_id: str):
```
ashwin1111, line 110: Modify all places where this function is being called, since there are change in parameters

ashwin1111 (changes requested):

ashwin1111 (approved):

PR #127, opened 2021-12-24
send customized memo to sage-intacct

Sravanksk:


apps/workspaces/serializers.py
```diff
@@ -42,6 +42,9 @@ def validate(self, data):
         :param data: Non-validated data
         :return: upserted general settings object
         """
+        if self.partial:
```
ashwin1111, line 45: Will this skip below two validations?
```
        if not data['auto_map_employees'] and data['auto_create_destination_entity']:
            raise serializers.ValidationError(
                'Cannot set auto_create_destination_entity value if auto map employees is disabled')
        if data['auto_map_employees'] == 'EMPLOYEE_CODE' and data['auto_create_destination_entity']:
            raise serializers.ValidationError('Cannot enable auto create destination entity for employee code')

ashwin1111 (approved):

PR #135, opened 2022-02-23
Support For Saving Location Entity mapping After Sage Intacct Connection

NileshPant1999:


apps/mappings/models.py
```diff
@@ -6,6 +6,21 @@
...
 
 
+class LocationEntityMapping(models.Model):
+    """
+    Location Entity Mapping
+    """
+    id = models.AutoField(primary_key=True)
+    location_entity_name = models.CharField(max_length=255, help_text='SageIntacct Location Entity Name')
+    country_name = models.CharField(max_length=255, help_text='SageIntacct Location Entities Country', null=True)
+    destination_id = models.CharField(max_length=255, help_text='SageIntacct location entity id')
```
ashwin1111, line 16: shouldn't this be `null=True`?
How are we handling top level ids?

apps/mappings/serializers.py
```diff
@@ -10,3 +10,11 @@ class GeneralMappingSerializer(serializers.ModelSerializer):
...
         model = GeneralMapping
         fields = '__all__'
+
+class LocationEntityMappingSerializer(serializers.ModelSerializer):
+    """
+    Location Entity Mappings group serializer
+    """
+    class Meta:
+        model = LocationEntityMapping
+        fields = '__all__'
```
ashwin1111, line 20: new line

apps/sage_intacct/helpers.py
```diff
@@ -56,7 +56,7 @@ def sync_dimensions(si_credentials: SageIntacctCredential, workspace_id: int, di
         dimensions = [
             'locations', 'customers', 'departments', 'projects', 'expense_payment_types',
             'classes', 'charge_card_accounts','payment_accounts', 'vendors', 'employees', 'accounts',
-            'expense_types', 'items', 'user_defined_dimensions'
+            'expense_types', 'items', 'user_defined_dimensions', 'location_entities'
```
ashwin1111, line 59: Why do we need to sync `location_entities` on every sync? We can just sync them once during post of sage intacct creds. Would be nice if we can use a post save signal.

apps/sage_intacct/utils.py
```diff
@@ -263,7 +263,11 @@ def sync_location_entities(self):
                 'attribute_type': 'LOCATION_ENTITY',
                 'display_name': 'location entity',
                 'value': location_entity['NAME'],
-                'destination_id': location_entity['LOCATIONID']
+                'destination_id': location_entity['LOCATIONID'],
+                'detail': {
+                    'country': location_entity['OPCOUNTRY'],
+                    'currency': location_entity['CURRENCY']
```
ashwin1111, line 269: Just a question, where will this be impacted?

apps/workspaces/views.py
```diff
@@ -65,7 +65,7 @@ def post(self, request):
 
             FyleCredential.objects.update_or_create(
                 refresh_token=auth_tokens.refresh_token,
-                workspace_id=workspace.id
+                workspace_id=workspace.id,
```
ashwin1111, line 68: remove `,`

apps/workspaces/views.py
```diff
@@ -166,6 +166,10 @@ def post(self, request, **kwargs):
             workspace.fyle_org_id = org_id
             workspace.save()
 
+            fyle_connector = FyleConnector(refresh_token=refresh_token, workspace_id=kwargs['workspace_id'])
+
+            cluster_domain = fyle_connector.get_cluster_domain()['cluster_domain']
```
ashwin1111, line 171: I don't see `cluster_domain` being used

ashwin1111:
@NileshPant1999 let's also try to add tests along with new features

apps/workspaces/views.py
NileshPant1999: will remove this, came from tax changes

apps/sage_intacct/utils.py
NileshPant1999: was trying to use both to check weather to disable the import tax or not will remove currency now as we are not using location entity with country set as null

apps/sage_intacct/helpers.py
NileshPant1999: will remove it from here

apps/mappings/models.py
NileshPant1999: I thought of adding a dummy a value like a dummy string but null will also work will change it to null=True

apps/mappings/signals.py
```diff
@@ -15,10 +15,26 @@
...
+@receiver(post_save, sender=LocationEntityMapping)
+def run_post_location_entity_mappings_triggers(sender, instance: LocationEntityMapping, **kwargs):
+    """
+    :param sender: Sender Class
+    :param instance: Row instance of Sender Class
+    :return: None
+    """
+
+    sage_intacct_credentials = SageIntacctCredential.objects.get(workspace_id=int(instance.workspace_id))
+    sync_dimensions(sage_intacct_credentials, int(instance.workspace_id))
```
ashwin1111, line 36: let's do this in UI, this will block client

apps/sage_intacct/utils.py
```diff
@@ -39,15 +39,20 @@ def __init__(self, credentials_object: SageIntacctCredential, workspace_id: int)
         decrypted_password = cipher_suite.decrypt(credentials_object.si_user_password.encode('utf-8')).decode('utf-8')
 
         # TODO: Cache general_mappings
-        general_mappings = GeneralMapping.objects.filter(workspace_id=workspace_id).first()
+        location_entity_mapping = LocationEntityMapping.objects.filter(workspace_id=workspace_id).first()
```
ashwin1111, line 42: add destination id shouldn't be 'top_level' in orm filter. And we can solve this in a single line
```
        self.connection = SageIntacctSDK(
            sender_id=sender_id,
            sender_password=sender_password,
            user_id=credentials_object.si_user_id,
            company_id=credentials_object.si_company_id,
            user_password=decrypted_password,
            entity_id=location_entity_mapping.destination_id if location_entity_mapping else None
        )

File: apps/workspaces/signals.py [hunk 9]
```diff
@@ -4,11 +4,12 @@
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
+from fyle_accounting_mappings.models import DestinationAttribute
```
ashwin1111, line 7: not needed

apps/workspaces/signals.py
```diff
@@ -4,11 +4,12 @@
...
 from apps.fyle.helpers import add_expense_id_to_expense_group_settings
 from apps.sage_intacct.helpers import schedule_payment_sync
 from apps.mappings.helpers import schedule_or_delete_auto_mapping_tasks
 
-from .models import Configuration
+from .models import Configuration, SageIntacctCredential
```
ashwin1111, line 12: not needed

ashwin1111 (approved):

PR #136, opened 2022-02-23
Migrate Cluster Domain To Fyle Credentials Table

NileshPant1999:


apps/workspaces/models.py
```diff
@@ -109,6 +109,7 @@ class FyleCredential(models.Model):
     """
     id = models.AutoField(primary_key=True)
     refresh_token = models.TextField(help_text='Stores Fyle refresh token')
+    cluster_domain = models.TextField(null=True, help_text='Cluster Domain')
```
ashwin1111, line 112: I remember we have cluster domain in other table too. Let's write a script to copy data from them. Only the unfilled ones, let's write python script. Also, we need to handle new orgs onboarding, is that taken care of?

NileshPant1999: yes onboarding new customer is taken care of in tax pr

ashwin1111 (approved):

PR #157, opened 2022-05-17
platform phase 3&4 migration

Sravanksk:


requirements.txt
```diff
@@ -9,8 +9,8 @@ django-rest-framework==0.1.0
 djangorestframework==3.11.2
 future==0.18.2
 fyle_accounting_mappings==1.7.0
-fyle-rest-auth==0.7.0
-fylesdk==2.3.1
```
ashwin1111, line 13: cross check across repo if we are using sdk

ashwin1111 (approved):

PR #158, opened 2022-05-17
Bug Fixes: platform phase-2 migrations

Sravanksk:


NileshPant1999 (approved):

apps/fyle/models.py
```diff
@@ -382,7 +382,6 @@ def create_or_update_reimbursement_objects(reimbursements: List[Dict], workspace
                         settlement_id=reimbursement['settlement_id'],
                         reimbursement_id=reimbursement['id'],
                         state=reimbursement['state'],
-                        payment_number=reimbursement['unique_id'],
```
ashwin1111, line 385: Why are we removing this? This is used in memo stuffs

apps/mappings/tasks.py
```diff
@@ -335,15 +335,17 @@ def auto_create_cost_center_mappings(workspace_id: int):
...
         )
 
+        platform = PlatformConnector(fyle_credentials=fyle_credentials)
+
         mapping_setting = MappingSetting.objects.get(
             source_field='COST_CENTER', import_to_fyle=True, workspace_id=workspace_id
         )
 
-        fyle_connection.sync_cost_centers()
+        platform.sync_cost_centers()
```
ashwin1111, line 344: I don't think `platform.sync_cost_centers` is right

ashwin1111 (changes requested):

ashwin1111 (approved):

PR #175, opened 2022-06-13
Changes to updated existing categories

NileshPant1999:


apps/mappings/tasks.py
```diff
@@ -228,7 +228,24 @@ def schedule_auto_map_charge_card_employees(workspace_id: int):
             schedule.delete()
 
 
-def create_fyle_categories_payload(categories: List[DestinationAttribute], workspace_id: int):
+def get_all_categories_from_fyle(platform: PlatformConnector):
+    categories_generator = platform.categories.get_all_generator()
```
Sravanksk, line 232: This will not return disabled categories - You should pull disabled categories too, so that you can enable again
Please fix in platform-connector or use platform SDK directly

PR #182, opened 2022-07-26
save category mappings to new Category MappingsTable

Sravanksk:


apps/mappings/tasks.py
```diff
@@ -586,78 +587,157 @@ def upload_categories_to_fyle(workspace_id: int, reimbursable_expenses_object: s
...
+        destination_id_pk_map[attribute['destination_id'].lower()] = attribute['id']
+
+    mapping_updation_batch = []
+
+    for category_mapping in category_mappings:
+        ccc_account_id = destination_id_pk_map[category_mapping.destination_expense_head.detail['account_internal_id'].lower()]
+        mapping_updation_batch.append(
+            CategoryMapping(
+                id=category_mapping.id,
+                source_category_id=category_mapping.source_category.id,
```
ashwin1111, line 628: Why is this needed? aren't we supposed to have a source category already?

ashwin1111: we are also not updating them, check line 635

apps/sage_intacct/tasks.py
```diff
@@ -432,33 +434,24 @@ def __validate_expense_group(expense_group: ExpenseGroup, configuration: Configu
...
-                source_type='CATEGORY',
-                source__value=category,
-                destination_type='CCC_ACCOUNT',
-                workspace_id=expense_group.workspace_id
-            ).first()
+        if account:
+            if configuration.reimbursable_expenses_object == 'EXPENSE_REPORT':
+                account = account.destination_expense_head
+            else:
+                account = account.destination_account
```
ashwin1111, line 446: If the org doesn't use CCC exports, we might end up blocking them with mapping errors. Let me know if it won't.

ashwin1111: Also, assume this is a CCC expense and config is ER -> CCT. Now, if the expense type exists for reimbursable type and gl account doesn't exist for ccc, we might end up allowing them.

These two cases needs handling.

ashwin1111:
Please assign back review once you fix the form validators for category

apps/mappings/tasks.py
ashwin1111, line 628: @Sravanksk this is not fixed yet

ashwin1111 (approved):

PR #183, opened 2022-07-28
Migrate Employeee mappings to new table

Sravanksk:


apps/sage_intacct/tasks.py
```diff
@@ -98,27 +101,51 @@ def create_or_update_employee_mapping(expense_group: ExpenseGroup, sage_intacct_
                      **filters
                 ).first()
 
-            if entity is None:
-                if employee_field_mapping == 'EMPLOYEE':
-                    entity: DestinationAttribute = sage_intacct_connection.get_or_create_employee(source_employee)
-                else:
+            existing_employee_mapping = EmployeeMapping.objects.filter(
```
ashwin1111, line 104: This comes in the `except EmployeeMapping.DoesNotExist:` block.
When do you think we would have employee mapping?

apps/sage_intacct/tasks.py
```diff
@@ -98,27 +101,51 @@ def create_or_update_employee_mapping(expense_group: ExpenseGroup, sage_intacct_
...
+                source_employee=source_employee
+            ).first()
+
+            destination = {}
+            if employee_field_mapping == 'EMPLOYEE':
+                if entity is None:
+                    if employee_field_mapping == 'EMPLOYEE':
```
ashwin1111, line 111: line 109 has the same check, why is this needed again?

apps/sage_intacct/tasks.py
```diff
@@ -98,27 +101,51 @@ def create_or_update_employee_mapping(expense_group: ExpenseGroup, sage_intacct_
...
+                        entity: DestinationAttribute = sage_intacct_connection.get_or_create_employee(source_employee)
+                        destination['destination_employee_id'] = entity.id
+                    elif existing_employee_mapping and existing_employee_mapping.destination_employee:
```
ashwin1111, line 114: same comment as line 104

apps/sage_intacct/tasks.py
```diff
@@ -98,27 +101,51 @@ def create_or_update_employee_mapping(expense_group: ExpenseGroup, sage_intacct_
...
+                        destination['destination_employee_id'] = existing_employee_mapping.destination_employee.id
+
+            else:
+                if entity is None:
                     entity: DestinationAttribute = sage_intacct_connection.get_or_create_vendor(
                         source_employee.detail['full_name'], source_employee.value, create=True
                     )
+                    destination['destination_vendor_id'] = entity.id
+                elif existing_employee_mapping and existing_employee_mapping.destination_vendor:
```
ashwin1111, line 123: same

apps/sage_intacct/tasks.py
```diff
@@ -403,23 +436,19 @@ def __validate_expense_group(expense_group: ExpenseGroup, configuration: Configu
...
             elif configuration.corporate_credit_card_expenses_object == 'JOURNAL_ENTRY':
                 error_message = 'Employee Mapping not found'
-                Mapping.objects.get(
-                    Q(destination_type='EMPLOYEE') | Q(destination_type='VENDOR'),
-                    source_type='EMPLOYEE',
-                    source__value=expense_group.description.get('employee_email'),
+                EmployeeMapping.objects.get(
+                    source_employee__value=expense_group.description.get('employee_email'),
                     workspace_id=expense_group.workspace_id
                 )
```
ashwin1111, line 442: we should check if destination card exist

apps/sage_intacct/tasks.py
```diff
@@ -403,23 +436,19 @@ def __validate_expense_group(expense_group: ExpenseGroup, configuration: Configu
...
                 error_message = 'Employee Mapping not found'
-                Mapping.objects.get(
-                    Q(destination_type='EMPLOYEE'),
-                    source_type='EMPLOYEE',
-                    source__value=expense_group.description.get('employee_email'),
+                EmployeeMapping.objects.get(
+                    source_employee__value=expense_group.description.get('employee_email'),
                     workspace_id=expense_group.workspace_id
-                )            
+                )
```
ashwin1111, line 449: we should check if destination employee exist

apps/sage_intacct/tasks.py
Sravanksk: destination card is not needed as we have a default ChargeCard in general mappings

apps/sage_intacct/tasks.py
Sravanksk: we have a check for this already in line 413

apps/sage_intacct/tasks.py
Sravanksk: This might come up during config updates, also in auto-map we fill employee_mappings table with source value, so this would exist at this point

ashwin1111: config update won't have an effect - can discuss on huddle tomo
also in auto-map we fill employee_mappings table with source value, so this would exist at this point - didn't understand this point, can discuss in same huddle

apps/sage_intacct/tasks.py
```diff
@@ -436,18 +435,33 @@ def __validate_expense_group(expense_group: ExpenseGroup, configuration: Configu
...
-                EmployeeMapping.objects.get(
+                entity = EmployeeMapping.objects.get(
                     source_employee__value=expense_group.description.get('employee_email'),
                     workspace_id=expense_group.workspace_id
                 )
 
+                if configuration.employee_field_mapping == 'EMPLOYEE':
+                    entity = entity.destination_employee
+                else:
+                    entity = entity.destination_vendor
```
ashwin1111, line 460: ER as CCC can be only for ER as reimburse, so can check for entity.destination_employee directly

ashwin1111 (approved):

PR #201, opened 2022-09-23
added unit tests for fyle module

harshithaputtaswamy:


fyle_intacct_api/tests/settings.py
```diff
@@ -212,12 +212,12 @@
         }
     }
 
-DATABASES['cache_db'] = {
-    'ENGINE': 'django.db.backends.sqlite3',
-    'NAME': 'cache.db'
-}
+# DATABASES['cache_db'] = {
+#     'ENGINE': 'django.db.backends.sqlite3',
```
NileshPant1999, line 216: why is this commented

tests/test_fyle/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+    
+    ExpenseGroupSettings.objects.create(
+        reimbursable_expense_group_fields='{employee_email,report_id,claim_number,fund_source}',
+        corporate_credit_card_expense_group_fields='{fund_source,employee_email,claim_number,expense_id,report_id}',
+        expense_state='PAYMENT PROCESSING',
+        workspace_id=98,
+        reimbursable_export_date_type='current_date',
+        ccc_export_date_type='spent_at'
+    )
+
```
NileshPant1999, line 31: fix this please

fyle_intacct_api/tests/settings.py
harshithaputtaswamy: while testing it will use sqlite if not commented

PR #207, opened 2022-09-27
add export type column in expense groups

Sravanksk:


apps/sage_intacct/tasks.py
```diff
@@ -551,6 +551,7 @@ def create_journal_entry(expense_group: ExpenseGroup, task_log_id):
             task_log.save()
 
             expense_group.exported_at = datetime.now()
+            expense_group.export_type = 'JOURNAL ENTRY'
```
ashwin1111, line 554: Just for consistence let's maintain _
`JOURNAL_ENTRY`
Same comment for all places

sql/scripts/015-generate-export-type-for-expense-groups.sql
```diff
@@ -0,0 +1,22 @@
+-- Script to add export_type to expense_groups table based on task_logs
+
+rollback;
+begin;
+
+update expense_groups set export_type = 'EXPENSE REPORT' where id in (
```
ashwin1111, line 6: same

ashwin1111 (approved):

PR #214, opened 2022-10-07
Change of Journal Entry behaviour - credit default charge card account, removed pylint

harshithaputtaswamy:


apps/sage_intacct/views.py
```diff
@@ -35,9 +36,19 @@ class DestinationAttributesView(generics.ListAPIView):
...
-            attribute_type__in=attribute_types, workspace_id=self.kwargs['workspace_id']).order_by('value')
+        destination_attributes_list = []
+
+        if 'BALANCESHEET_GL_ACCOUNT' in attribute_types:
+            attribute_types.remove('BALANCESHEET_GL_ACCOUNT')
+            destination_attributes = DestinationAttribute.objects.filter(
+                attribute_type__in=attribute_types, workspace_id=self.kwargs['workspace_id']).order_by('value')
+            gl_accounts = DestinationAttribute.objects.filter(
+                attribute_type='ACCOUNT', detail__contains={'account_type': 'balancesheet'}, workspace_id=self.kwargs['workspace_id']).order_by('value')
+            destination_attributes_list = list(chain(destination_attributes, gl_accounts))
```
Shwetabhk, line 47: there is no need to use chain
can just do  
```python
destination_attributes.extend(gl_accounts)
return destination_attributes
````

File: apps/sage_intacct/views.py [hunk 2]
```diff
@@ -1,3 +1,4 @@
+import re
```
ashwin1111, line 1: remove unused imports

apps/sage_intacct/views.py
```diff
@@ -35,9 +36,23 @@ class DestinationAttributesView(generics.ListAPIView):
 
     def get_queryset(self):
         attribute_types = self.request.query_params.get('attribute_types').split(',')
-
-        return DestinationAttribute.objects.filter(
-            attribute_type__in=attribute_types, workspace_id=self.kwargs['workspace_id']).order_by('value')
+        account_type = self.request.query_params.get('account_type')
```
ashwin1111, line 39: If we do this, we don't have to filter something and later exclude it. Also sending balance sheet from client and filtering !incomestatement isn't very appropriate. Later if intacct introduces some other type we might be in trouble.

```
# Send account_type as incomestatement from client

account_type = self.request.query_params.get('account_type', None)

params = {
    'attribute_type__in': attribute_types,
    'workspace_id': self.kwargs['workspace_id']
}

if account_type:
    # Also, instead of detail__contains check if detail__account_type is working -> this shouldn't break for other attributes where detail column is null
    params['detail__contains'] = {
        'account_type': account_type
    }

return DestinationAttribute.objects.filter(**params).order_by('value')

Shwetabhk (approved):

PR #218, opened 2022-12-09
fixed auto_now to auto_now_add for created_at fields

vishalpandeyvip:
created_at field of APPayment and SageIntacctReimbursement had auto_now = True, that will update created_at field everytime save method will be called on their object, hence they are changed to auto_now_add = True


Shwetabhk (dismissed):

apps/sage_intacct/models.py
```diff
@@ -1117,7 +1117,7 @@ class SageIntacctReimbursement(models.Model):
     employee_id = models.CharField(max_length=255, help_text='Sage Intacct Employee ID')
     memo = models.TextField(help_text='Reimbursement Memo')
     payment_description = models.TextField(help_text='Reimbursement Description')
-    created_at = models.DateField(auto_now=True, help_text='Created at')
+    created_at = models.DateField(auto_now_add=True, help_text='Created at')
```
Shwetabhk, line 1120: Migrations are not added

Shwetabhk (approved):

PR #229, opened 2023-01-23
Auto sync of projects and chart of accounts

labhvam5:


apps/sage_intacct/utils.py
```diff
@@ -91,23 +91,46 @@ def sync_accounts(self):
...
         Get accounts
         """
-        accounts = self.connection.accounts.get_all(field='STATUS', value='active')
+        accounts = self.connection.accounts.get_all()
 
         account_attributes = {
             'account': [],
             'ccc_account': []
         }
+        destination_attributes = DestinationAttribute.objects.filter(workspace_id=self.workspace_id,\
```
Shwetabhk, line 100: We don't need `\` for function params, you can just change the line and indent correctly

apps/sage_intacct/utils.py
```diff
@@ -91,23 +91,46 @@ def sync_accounts(self):
...
+                    'destination_id': account['ACCOUNTNO'],
+                    'active': True,
+                    'detail': {
+                        'account_type': account['ACCOUNTTYPE']
+                    }
+                })
+                if account['ACCOUNTNO'] in disabled_fields_map:
+                    disabled_fields_map.pop(account['ACCOUNTNO'])
+
+        #For setting active to False
```
Shwetabhk, line 125:
```suggestion
        # For setting active to False
```

apps/sage_intacct/utils.py
```diff
@@ -91,23 +91,46 @@ def sync_accounts(self):
...
+        for destination_id in disabled_fields_map:
```
Shwetabhk, line 126: I think deletion is not handled, what if the account is deleted from Sage Intacct?

Shwetabhk: This will only disable the ones which are in the `disabled_field_maps`

What about the ones which are not returned at all by the API? Have we confirmed that that can't be a case?

apps/sage_intacct/utils.py
```diff
@@ -141,21 +164,46 @@ def sync_expense_types(self):
         """
         Get expense types
         """
-        expense_types = self.connection.expense_types.get_all(field='STATUS', value='active')
+        expense_types = self.connection.expense_types.get_all()
 
         expense_types_attributes = []
+        destination_attributes = DestinationAttribute.objects.filter(workspace_id=self.workspace_id,\
```
Shwetabhk, line 170: Same comments as above for this function

apps/sage_intacct/utils.py
```diff
@@ -210,9 +258,18 @@ def sync_projects(self):
         """
         projects_count = self.connection.projects.count()
         if projects_count < SYNC_UPPER_LIMIT['projects']:
-            projects = self.connection.projects.get_all(field='STATUS', value='active')
+            projects = self.connection.projects.get_all()
 
             project_attributes = []
+            destination_attributes = DestinationAttribute.objects.filter(workspace_id=self.workspace_id,\
```
Shwetabhk, line 264: Same comments as top function

apps/sage_intacct/utils.py
```diff
@@ -229,6 +286,19 @@ def sync_projects(self):
                         'active': True,
                         'detail': detail
                     })
+                    if project['PROJECTID'] in disabled_fields_map:
+                        disabled_fields_map.pop(project['PROJECTID'])
+
+            #For setting active to False
```
Shwetabhk, line 292:
```suggestion
            # For setting active to False
```

Shwetabhk (changes requested):

apps/mappings/tasks.py
```diff
@@ -777,12 +858,24 @@ def auto_create_category_mappings(workspace_id):
...
             workspace_id=workspace_id, reimbursable_expenses_object=reimbursable_expenses_object,
             corporate_credit_card_expenses_object=corporate_credit_card_expenses_object)
 
         create_category_mappings(fyle_categories, reimbursable_destination_type, workspace_id)
 
+        # auto-sync categories and expense accounts
+        category_ids_to_be_changed = disable_expense_attributes('CATEGORY', reimbursable_destination_type, workspace_id)
+        if category_ids_to_be_changed:
+            expense_attributes = ExpenseAttribute.objects.filter(id__in=category_ids_to_be_changed)
+            fyle_payload: List[Dict] = create_fyle_categories_payload(categories=[], category_map={}, updated_categories=expense_attributes, destination_type=reimbursable_destination_type)
```
NileshPant1999, line 874: can pass fyle credentials in this to save one extra query

labhvam5: We dont use `FyleCredentials` in the `create_fyle_categories_payload` function .

NileshPant1999: ohh sorry i mean in upload_categories to fyle function

apps/mappings/tasks.py
```diff
@@ -48,23 +102,36 @@ def create_fyle_projects_payload(projects: List[DestinationAttribute], existing_
...
+                'code': destination_id_of_project,
+                'description': 'Project - {0}, Id - {1}'.format(
                     project.value,
-                    project.destination_id
+                    destination_id_of_project
                 ),
-                'is_enabled': True if project.active is None else project.active
+                'is_enabled': project.active
             })
+    else:
```
ashwin1111, line 118: move line 103 to this block

apps/sage_intacct/utils.py
```diff
@@ -141,21 +168,48 @@ def sync_expense_types(self):
...
         """
-        expense_types = self.connection.expense_types.get_all(field='STATUS', value='active')
+        expense_types = self.connection.expense_types.get_all()
 
         expense_types_attributes = []
+        destination_attributes = DestinationAttribute.objects.filter(workspace_id=self.workspace_id,
+                attribute_type= 'EXPENSE_TYPE', display_name='Expense Types').values('destination_id', 'value', 'detail')
+        disabled_fields_map = {}
+
+        for destination_attribute in destination_attributes:
```
ashwin1111, line 178: most of the logic seems to be same, please separate out as common function for disabling operation

apps/sage_intacct/utils.py
```diff
@@ -229,6 +292,23 @@ def sync_projects(self):
...
+                # If yes then we pop the item from the disable_field_map else we set the active = True.
+            # This should take care of delete as well as inactive case since we are checking the status=Active case.
+            for destination_id in disabled_fields_map:
+                project_attributes.append({
+                    'attribute_type': 'PROJECT',
+                    'display_name': 'project',
+                    'value': disabled_fields_map[destination_id]['value'],
+                    'destination_id': destination_id,
+                    'active': False,
+                    'detail': disabled_fields_map[destination_id]['detail']
```
ashwin1111, line 310: sync_customers is not handled

apps/sage_intacct/utils.py
labhvam5: will do it during rewrite can skip for now.

apps/sage_intacct/utils.py
labhvam5: We dont need to add it for `customers` we are only disabling in case of `Project <> Project` mappings.

ashwin1111: arey customer is also Project, check attribute type

ashwin1111: replied to other comment

ashwin1111 (approved):

Shwetabhk (approved):

PR #235, opened 2023-02-07
export refunds as bills and journal entry feature

vishalpandeyvip:


apps/sage_intacct/utils.py
```diff
@@ -1025,6 +1025,15 @@ def __construct_journal_entry(self, journal_entry: JournalEntry, journal_entry_l
                 }
             }
 
+            refund = lineitem.amount < 0
+            if refund:
```
labhvam5, line 1029: you can just add `if lineitem.amount < 0:`

vishalpandeyvip: yes, that could be done, but I think this code is more descriptive and explains why this block of code is required.

labhvam5 (approved):

ashwin1111 (approved):

Shwetabhk (approved):

NileshPant1999 (approved):

PR #242, opened 2023-03-02
Cost Code and Cost Type as Dependent Fields In Fyle

NileshPant1999:


ashwin1111:
will review once the tests pass

.github/workflows/pytest.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         run: |
           docker-compose -f docker-compose-pipeline.yml build
           docker-compose -f docker-compose-pipeline.yml up -d
-          docker-compose -f docker-compose-pipeline.yml exec -T api pytest tests/ --cov --junit-xml=test-reports/report.xml --cov-report=term-missing --cov-fail-under=98 | tee pytest-coverage.txt
+          docker-compose -f docker-compose-pipeline.yml exec -T api pytest tests/ --cov --junit-xml=test-reports/report.xml --cov-report=term-missing --cov-fail-under=96 | tee pytest-coverage.txt
```
ashwin1111, line 19: 🔪

apps/mappings/tasks.py
```diff
@@ -482,7 +488,7 @@ def schedule_cost_centers_creation(import_to_fyle, workspace_id):
 
 
 def create_fyle_expense_custom_field_payload(sageintacct_attributes: List[DestinationAttribute], workspace_id: int,
-                                             fyle_attribute: str, platform: PlatformConnector):
+                                             fyle_attribute: str, platform: PlatformConnector, parent_field: str = None):
```
ashwin1111, line 491: name it `parent_field_id`. Also I assume this is stored as str in Fyle DB. If it's stored as int, we should also do the same.

apps/mappings/tasks.py
```diff
@@ -505,15 +511,31 @@ def create_fyle_expense_custom_field_payload(sageintacct_attributes: List[Destin
...
-            'is_mandatory': False,
-            'placeholder': 'Select {0}'.format(fyle_attribute),
-            'options': fyle_expense_custom_field_options,
-            'code': None
-        }
+        # if parent field is there that means it is a dependent field
+        if parent_field:
+            expense_custom_field_payload = {
+                'field_name': fyle_attribute,
+                'column_name': fyle_attribute,
```
ashwin1111, line 518: is this a mandatory key? https://docs.fylehq.com/docs/fyle-platform-docs/3529a1c92029b-create-or-update-an-expense-fields.

apps/mappings/tasks.py
```diff
@@ -505,15 +511,31 @@ def create_fyle_expense_custom_field_payload(sageintacct_attributes: List[Destin
...
+                'type': 'DEPENDENT_SELECT',
+                'is_custom': True,
+                'is_enabled': True,
+                'is_mandatory': False,
+                'placeholder': 'Select {0}'.format(fyle_attribute),
```
ashwin1111, line 523: We should preserve existing placeholder, remember Rushikesh fixed it somewhere

apps/mappings/tasks.py
```diff
@@ -505,15 +511,31 @@ def create_fyle_expense_custom_field_payload(sageintacct_attributes: List[Destin
...
+                'parent_field_id': parent_field,
+                'code': None,
+            }
+        else:
+            expense_custom_field_payload = {
+                'field_name': fyle_attribute,
+                'type': 'SELECT',
+                'is_enabled': True,
+                'is_mandatory': False,
+                'placeholder': 'Select {0}'.format(fyle_attribute),
```
ashwin1111, line 534: We should preserve existing placeholder, remember Rushikesh fixed it somewhere

apps/mappings/tasks.py
```diff
@@ -523,7 +545,58 @@ def create_fyle_expense_custom_field_payload(sageintacct_attributes: List[Destin
...
 
-def upload_attributes_to_fyle(workspace_id: int, sageintacct_attribute_type: str, fyle_attribute_type: str):
+
+def upload_dependent_field_to_fyle(workspace_id: int, sageintacct_attribute_type: str, fyle_attribute_type: str, parent_field_id: str):
+    """
+    Upload Dependent Fields To Fyle
+    """
+
+    fyle_credentials: FyleCredential = FyleCredential.objects.get(workspace_id=workspace_id)
+    platform = PlatformConnector(fyle_credentials=fyle_credentials)
```
ashwin1111, line 555: are we not syncing intacct fields before we upload custom field?

apps/mappings/tasks.py
```diff
@@ -523,7 +545,58 @@ def create_fyle_expense_custom_field_payload(sageintacct_attributes: List[Destin
...
+    expense_field_id = ExpenseAttribute.objects.filter(
+        workspace_id=workspace_id, attribute_type=fyle_attribute_type
+    ).first().detail['custom_field_id']
+
+    sage_intacct_attributes: List[DestinationAttribute] = DestinationAttribute.objects.filter(
+        workspace_id=workspace_id, attribute_type=sageintacct_attribute_type
+    )
+    sage_intacct_attributes = remove_duplicates(sage_intacct_attributes, parent_field_id)
+
+    if sageintacct_attribute_type != 'TASK':
```
ashwin1111, line 567: why is this done? not able to understand

apps/mappings/tasks.py
```diff
@@ -523,7 +545,58 @@ def create_fyle_expense_custom_field_payload(sageintacct_attributes: List[Destin
...
+        if attribute.attribute_type != 'TASK':
+            # parent value is combination of these two so filterig it out
+            parent_expense_field_value = parent_field.filter(
+                detail__project_name=attribute.detail['project_name'],
+                detail__external_id=attribute.detail['task_id']
+            ).first().value
+
+        payload = {
+            "parent_expense_field_id": parent_field_id,
+            "parent_expense_field_value": attribute.detail['project_name'] if attribute.attribute_type == 'TASK' else parent_expense_field_value,
```
ashwin1111, line 585: for task as child, are we blindly believing project should it's parent?

apps/mappings/tasks.py
```diff
@@ -26,15 +26,21 @@
...
 
-    for attribute in si_attributes:
-        if attribute.value.lower() not in attribute_values:
-            unique_attributes.append(attribute)
-            attribute_values.append(attribute.value.lower())
+    if is_dependent:
+        for attribute in si_attributes:
+            if {attribute.detail['project_name']: attribute.value.lower()} not in attribute_values:
+                unique_attributes.append(attribute)
+                attribute_values.append({attribute.detail['project_name']: attribute.value.lower()})
```
ashwin1111, line 37: when we allow it for other types, we should explicitly fix all these

ashwin1111: write a generic code / add a todo

apps/sage_intacct/utils.py
```diff
@@ -258,6 +258,61 @@ def sync_payment_accounts(self):
 
         return []
 
+    def sync_tasks(self):
+        """
+        Get of Tasks
+        """
+
+        intacct_tasks = self.connection.tasks.get_all()
```
ashwin1111, line 266: let' add active filter please

apps/sage_intacct/utils.py
```diff
@@ -258,6 +258,61 @@ def sync_payment_accounts(self):
...
+        task_attributes = []
+
+        # saving values as combination of taskid, name and recordno to avoid duplicates 
+        for task in intacct_tasks:
+            task_attributes.append({
+                'attribute_type': 'TASK',
+                'display_name': 'tasks',
```
ashwin1111, line 273: display name should be similar to attribute type

apps/sage_intacct/utils.py
```diff
@@ -258,6 +258,61 @@ def sync_payment_accounts(self):
...
+                'value': task['TASKID'] + '--' + task['NAME'] + '--' + task['RECORDNO'],
```
ashwin1111, line 274: use '{} -- {}'.format() for consistency

apps/sage_intacct/utils.py
```diff
@@ -258,6 +258,61 @@ def sync_payment_accounts(self):
...
+                'destination_id': task['RECORDNO'], # storing record number instead of internal id to avoid duplicates
```
ashwin1111, line 275: `# instead of TASKID*`

apps/mappings/tasks.py
NileshPant1999: sure will add this.

apps/mappings/tasks.py
NileshPant1999: adding a TODO for now

apps/mappings/tasks.py
NileshPant1999: we do @ashwin1111 in `auto_create_expense_fields_mappings` function

apps/mappings/tasks.py
ashwin1111: What about allowed values? This can be dynamic right?

apps/mappings/tasks.py
ashwin1111: what happens for the first run via signal?

apps/mappings/tasks.py
NileshPant1999: need to check this, but i remember seeing an error that field name and column name should be same will get this confirmed

apps/mappings/tasks.py
NileshPant1999: @ashwin1111 the behaviour is same as of custom field import.

apps/mappings/tasks.py
NileshPant1999: for sage intacct beta release yes already discussed this with Gayathiri

apps/mappings/tasks.py
ashwin1111: why not fix the gap? p2 though

.github/workflows/pytest.yml
NileshPant1999: creating tests in new pr will add before merging this, will maintain 98 only

apps/mappings/tasks.py
NileshPant1999: adding a todo for now

apps/mappings/tasks.py
NileshPant1999: make sense removing

apps/sage_intacct/utils.py
```diff
@@ -258,6 +258,61 @@ def sync_payment_accounts(self):
...
+        """
+        Get of Tasks
+        """
+
+        intacct_tasks = self.connection.tasks.get_all()
+        task_attributes = []
+
+        # saving values as combination of taskid, name and recordno to avoid duplicates 
+        for task in intacct_tasks:
+            task_attributes.append({
```
Shwetabhk, line 271: why are we not storing `active` field correctly?

apps/sage_intacct/utils.py
```diff
@@ -258,6 +258,61 @@ def sync_payment_accounts(self):
...
+
+        cost_types = self.connection.cost_types.get_all()
+        cost_types_attributes = []
+
+        for cost_type in cost_types:
+            cost_types_attributes.append({
+                'attribute_type': 'COST_TYPE',
+                'display_name': 'cost type',
+                'value': '{}--{}--{}'.format(cost_type['COSTTYPEID'], cost_type['NAME'], cost_type['RECORDNO']),
+                'destination_id': cost_type['RECORDNO'],
```
Shwetabhk, line 301: No `active` field

.github/workflows/pytest.yml
NileshPant1999: @Shwetabhk only did this as Gayathri want this on staging quickly for recording

ashwin1111: test is not a blocker for pvt branch deploys no?

apps/sage_intacct/utils.py
NileshPant1999: not available for this type

PR #251, opened 2023-03-22
Invalid token, Invalid Credentials, No attribute Date, Local Variable…

ruuushhh:
… Admin name errors resolved

### DIFF
--- a/apps/mappings/tasks.py
+++ b/apps/mappings/tasks.py
@@ -753,21 +753,18 @@ def upload_categories_to_fyle(workspace_id: int, reimbursable_expenses_object: s
     """
     Upload categories to Fyle
     """
-    try:
-        si_credentials: SageIntacctCredential = SageIntacctCredential.objects.get(workspace_id=workspace_id)
+    si_credentials: SageIntacctCredential = SageIntacctCredential.objects.get(workspace_id=workspace_id)
 
-        platform = PlatformConnector(fyle_credentials)
+    platform = PlatformConnector(fyle_credentials)
 
-        category_map = get_all_categories_from_fyle(platform=platform)
+    category_map = get_all_categories_from_fyle(platform=platform)
 
-    
-        si_connection = SageIntacctConnector(
-            credentials_object=si_credentials,
-            workspace_id=workspace_id
-        )
-        platform.categories.sync()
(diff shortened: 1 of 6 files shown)


apps/workspaces/tasks.py
```diff
@@ -162,14 +170,16 @@ def run_email_notification(workspace_id):
...
+            admin_name = 'Admin'
             if task_logs and (ws_schedule.error_count is None or len(task_logs) > ws_schedule.error_count):
                 context = {
                     'name': admin_name,
                     'errors': len(task_logs),
                     'fyle_company': workspace.name,
                     'intacct_company': intacct.si_company_name,
                     'workspace_id': workspace_id,
-                    'export_time': workspace.last_synced_at.date(),
+                    'export_time':  max(workspace.last_synced_at.date(), workspace.ccc_last_synced_at.date()) if workspace.last_synced_at.date() and workspace.ccc_last_synced_at.date() \
```
ashwin1111, line 181: move it out of the dict, looks complicated to read.

apps/workspaces/tasks.py
```diff
@@ -162,14 +170,16 @@ def run_email_notification(workspace_id):
                     if data['email'] == admin_email:
                         admin_name = data['name']
 
+            admin_name = 'Admin'
```
ashwin1111, line 173: this should be before the assignment, in line 164, else we end up updating admin_name to "Admin"

ashwin1111 (commented):
async_auto_map_employees(), auto_create_cost_center_mappings(), async_auto_create_custom_field_mappings(), and few other functions in same file  - SageIntacctCredential.DoesNotExist is missing

apps/mappings/tasks.py
```diff
@@ -753,17 +753,22 @@ def upload_categories_to_fyle(workspace_id: int, reimbursable_expenses_object: s
     """
     Upload categories to Fyle
     """
-    si_credentials: SageIntacctCredential = SageIntacctCredential.objects.get(workspace_id=workspace_id)
+    try:
```
ashwin1111, line 756: this is not needed, the function that calls this function is following a try catch pattern

ashwin1111 (approved):

ashwin1111 (approved):

PR #254, opened 2023-03-28
Platform connector error resolved

ruuushhh:


apps/mappings/tasks.py
```diff
@@ -188,8 +188,9 @@ def auto_create_project_mappings(workspace_id: int):
 
         post_projects_in_batches(platform, workspace_id, mapping_setting.destination_field)
 
-    except (SageIntacctCredential.DoesNotExist, InvalidTokenError):
-        logger.info('Invalid Token or Sage Intacct credentials does not exist - %s', workspace_id)
+    except (SageIntacctCredential.DoesNotExist, FyleCredential.DoesNotExist, InvalidTokenError):
```
ashwin1111, line 191: `FyleCredential.DoesNotExist` is not needed across this PR. This is an outdated error and this happened in async_update_fyle_credentials(). There wouldn't be a case where Fyle Credential doesn't exist for an org.

apps/mappings/tasks.py
```diff
@@ -188,8 +188,9 @@ def auto_create_project_mappings(workspace_id: int):
...
+        logger.info('Invalid Token or Sage/Fyle Intacct credentials does not exist - %s', workspace_id)
```
ashwin1111, line 192: after FC removal, update this messaging as well

apps/mappings/tasks.py
```diff
@@ -230,8 +231,8 @@ def async_auto_map_employees(workspace_id: int):
             sage_intacct_connection.sync_vendors()
 
         EmployeesAutoMappingHelper(workspace_id, destination_type, employee_mapping_preference).reimburse_mapping()
-    except (SageIntacctCredential.DoesNotExist, InvalidTokenError):
-        logger.info('Invalid Token or Sage Intacct Credentials does not exist - %s', workspace_id)
+    except (SageIntacctCredential.DoesNotExist, FyleCredential.DoesNotExist, InvalidTokenError):
```
ashwin1111, line 234: We have same exception name `InvalidTokenError` in both platform connector as well as Sage Intacct SDK. Create an alias and handle it.
Group exception into 2 categories, SageIntacctCredential.DoesNotExist / InvalidTokenError (from intacct)
and 
InvalidTokenError (from platform SDK)
update loggers too

apps/sage_intacct/tasks.py
```diff
@@ -55,6 +55,9 @@ def load_attachments(sage_intacct_connection: SageIntacctConnector, key: str, ex
 
         supdoc_id = key
         return sage_intacct_connection.post_attachments(attachments, supdoc_id)
+    except FyleCredential.DoesNotExist:
```
ashwin1111, line 58: not needed

ashwin1111 (approved):

PR #256, opened 2023-04-06
5xx error handled and handle module permission exception handled

ruuushhh:


apps/mappings/tasks.py
```diff
@@ -194,6 +194,8 @@ def auto_create_project_mappings(workspace_id: int):
     except FyleInvalidTokenError:
         logger.info('Invalid Token for fyle')
     
+    except InternalServerError:
+        logger.error('Internal server error fyle')
```
ashwin1111, line 198: Internal server error while importing to Fyle

apps/mappings/tasks.py
```diff
@@ -240,6 +242,9 @@ def async_auto_map_employees(workspace_id: int):
     except FyleInvalidTokenError:
         logger.info('Invalid Token for fyle')
 
+    except NoPrivilegeError:
+        logger.info('Insufficient permissions')
```
ashwin1111, line 246: Insufficient permission to access the requested module

apps/sage_intacct/tasks.py
```diff
@@ -643,6 +643,9 @@ def create_journal_entry(expense_group: ExpenseGroup, task_log_id):
     except InvalidTokenError as exception:
         handle_sage_intacct_errors(exception, expense_group, task_log, 'Journal Entry')
 
+    except NoPrivilegeError:
```
ashwin1111, line 646: club all exceptions that calls handle_sage_intacct_errors() in this file

apps/sage_intacct/tasks.py
```diff
@@ -1271,6 +1297,10 @@ def check_sage_intacct_object_status(workspace_id):
         logger.info('Invalid Token or SageIntacct credentials does not exist - %s', workspace_id)
         return 
 
+    except NoPrivilegeError:
```
ashwin1111, line 1300: can club this to above exception, update message

apps/sage_intacct/tasks.py
ruuushhh: Can you explain this a bit?

ashwin1111: wherever we're calling handle_sage_intacct_errors(), you can add this exception to that list. Example - 
```
except (InvalidTokenError, NoPrivilegeError) as exception:
        handle_sage_intacct_errors(exception, expense_group, task_log, 'Charge Card Transactions')

ashwin1111 (approved):

PR #258, opened 2023-04-10
Intacct export memo changes

ruuushhh:


apps/sage_intacct/models.py
```diff
@@ -269,7 +270,7 @@ def get_transaction_date(expense_group: ExpenseGroup) -> str:
     return datetime.now().strftime('%Y-%m-%dT%H:%M:%S')
 
 
-def get_memo(expense_group: ExpenseGroup, payment_type: str=None) -> str:
+def get_memo(expense_group: ExpenseGroup, payment_type: str=None, export_table: Union['Bill', 'ExpenseReport', 'JournalEntry', 'ChargeCardTransaction', 'APPayment', 'SageIntacctReimbursement']=None) -> str:
```
ashwin1111, line 273: we don't need = None for export table. Also, let's name in ExportTable to make it feel like a model.

apps/sage_intacct/models.py
```diff
@@ -304,15 +306,23 @@ def get_memo(expense_group: ExpenseGroup, payment_type: str=None) -> str:
             workspace_id=expense_group.workspace_id
         )
         if expense_group.fund_source == 'CCC':
+            if export_table:
+                count = export_table.objects.filter(memo__contains = memo).count()
```
ashwin1111, line 310: do this outside the 308 if block.

apps/sage_intacct/models.py
```diff
@@ -304,15 +306,23 @@ def get_memo(expense_group: ExpenseGroup, payment_type: str=None) -> str:
...
             if expense_group_settings.ccc_export_date_type != 'current_date':
                 date = get_transaction_date(expense_group)
                 date = (datetime.strptime(date, '%Y-%m-%dT%H:%M:%S')).strftime('%d/%m/%Y')
                 memo = '{} - {}'.format(memo, date)
+            if count > 0:
```
ashwin1111, line 315: do this once after the if else block, after 325 since we're doing the same.

apps/sage_intacct/models.py
```diff
@@ -303,6 +305,8 @@ def get_memo(expense_group: ExpenseGroup, payment_type: str=None) -> str:
         expense_group_settings: ExpenseGroupSettings = ExpenseGroupSettings.objects.get(
             workspace_id=expense_group.workspace_id
         )
+        if ExportTable:
```
ashwin1111, line 308: update this condition to `if not payment_type`. We will send `ExportTable` every time, so this check is not relevant. But, we shouldn't be doing this for payment sync runs.

apps/sage_intacct/models.py
```diff
@@ -303,6 +305,8 @@ def get_memo(expense_group: ExpenseGroup, payment_type: str=None) -> str:
...
+                count = ExportTable.objects.filter(memo__contains = memo).count()
```
ashwin1111, line 309: `memo__contains=memo` no space before and after =

ashwin1111 (approved):

PR #268, opened 2023-04-13
reimbursable Journal Entry

anishfyle:


apps/sage_intacct/utils.py
```diff
@@ -1036,7 +1036,7 @@ def __construct_journal_entry(self, journal_entry: JournalEntry, journal_entry_l
         for lineitem in journal_entry_lineitems:
             expense_link = self.get_expense_link(lineitem)
             credit_line = {
-                'accountno': general_mappings.default_credit_card_id,
+                'accountno': general_mappings.default_credit_card_id if general_mappings.default_credit_card_id else general_mappings.default_gl_account_id,
```
Shwetabhk, line 1039: this is not right, if I am using both credit card and reimbursable as journal, we'll always be selecting the default_credit_card_id

we should computing this based on fund_source

anishfyle: yes @Shwetabhk, will fix this.

apps/sage_intacct/utils.py
```diff
@@ -1063,6 +1063,35 @@ def __construct_journal_entry(self, journal_entry: JournalEntry, journal_entry_l
                 }
             }
 
+        if configuration.reimbursable_expenses_object == 'JOURNAL_ENTRY':
+            credit_line_re = {
```
ashwin1111, line 1067: separate dict is not needed, can just update the existing dict's accountno key.

apps/sage_intacct/utils.py
```diff
@@ -1063,6 +1063,35 @@ def __construct_journal_entry(self, journal_entry: JournalEntry, journal_entry_l
...
+                'accountno': general_mappings.default_gl_account_id,
```
ashwin1111, line 1068: `'accountno': general_mappings.default_credit_card_id if journal_entry.expense_group.fund_source == 'CCC' else general_mappings.default_gl_account_id`

apps/sage_intacct/utils.py
```diff
@@ -1110,9 +1139,11 @@ def __construct_journal_entry(self, journal_entry: JournalEntry, journal_entry_l
             for dimension in lineitem.user_defined_dimensions:
                 for name, value in dimension.items():
                     credit_line[name] = value
+                    credit_line_re[name] = value
```
ashwin1111, line 1142: not needed

apps/sage_intacct/utils.py
```diff
@@ -1110,9 +1139,11 @@ def __construct_journal_entry(self, journal_entry: JournalEntry, journal_entry_l
...
                     debit_line[name] = value
 
             journal_entry_payload.append(credit_line)
+            journal_entry_payload.append(credit_line_re)
```
ashwin1111, line 1146: same

Shwetabhk (approved):

PR #284, opened 2023-05-16
export item to sage intacct

DhaaraniCIT:


apps/sage_intacct/models.py
```diff
@@ -152,14 +152,35 @@ def get_customer_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
+        if mapping:
+            item_id = mapping.destination.destination_id
+        else:
+            project_setting: MappingSetting = MappingSetting.objects.filter(
+                workspace_id=expense_group.workspace_id,
+                destination_field='PROJECT'
+            ).first()
+            item_id = general_mappings.default_item_id if general_mappings.default_item_id and project_setting else None
+    else:
         project_setting: MappingSetting = MappingSetting.objects.filter(
```
ashwin1111, line 179: this is not needed now

ashwin1111 (approved):

apps/sage_intacct/models.py
```diff
@@ -152,14 +152,33 @@ def get_customer_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
+    item_setting: MappingSetting = MappingSetting.objects.filter(
+        workspace_id=expense_group.workspace_id,
+        destination_field='ITEM'
+    ).first()
+
+    if item_setting:
+        attribute = ExpenseAttribute.objects.filter(attribute_type=item_setting.source_field).first()
+        source_value = lineitem.custom_properties.get(attribute.display_name, None)
+        if source_value is None:
+            lineitem = lineitem.__dict__
```
ashwin1111, line 164: why are we doing this? we're doing it anywhere

apps/sage_intacct/models.py
```diff
@@ -152,14 +152,33 @@ def get_customer_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
+            if item_setting.source_field == 'MERCHANT' or item_setting.source_field == 'VENDOR':
+                source_value = lineitem.vendor
```
ashwin1111, line 166: this mapping cannot exist, you won't be able to map merchant / vendor to item in the 1st place

apps/sage_intacct/models.py
```diff
@@ -152,14 +152,33 @@ def get_customer_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
+            elif lineitem[attribute.display_name.replace(" ","_").lower()] is not None:
+                source_value = lineitem[attribute.display_name.replace(" ","_").lower()]
```
ashwin1111, line 168: why all these hacks? can we follow what we do for other fields?

apps/sage_intacct/models.py
```diff
@@ -152,14 +152,33 @@ def get_customer_id_or_none(expense_group: ExpenseGroup, lineitem: Expense, gene
...
+            workspace_id=expense_group.workspace_id
         ).first()
-        if project_setting:
+        if mapping:
+            item_id = mapping.destination.destination_id
+        else:
             item_id = general_mappings.default_item_id if general_mappings.default_item_id else None
-
+    else:
+        item_id = general_mappings.default_item_id if general_mappings.default_item_id else None
```
ashwin1111, line 181: `general_mappings.default_item_id if general_mappings.default_item_id else None` this logic seems to be duplicated, do this at last when you don't have an item id

ashwin1111 (changes requested):

apps/sage_intacct/models.py
DhaaraniCIT: we can, i saw option of merchant while creating expense field

ashwin1111 (approved):

PR #292, opened 2023-05-23
AP payment bug resolved

ruuushhh:


Shwetabhk (approved):

apps/sage_intacct/tasks.py
```diff
@@ -1097,7 +1098,7 @@ def schedule_ap_payment_creation(sync_fyle_to_sage_intacct_payments, workspace_i
                 }
             )
 
-        if not sync_fyle_to_sage_intacct_payments:
+        if not sync_fyle_to_sage_intacct_payments or configuration.reimbursable_expenses_object != 'BILL':
```
Shwetabhk, line 1101: this should not be `or` it should be `and`

Shwetabhk: in case of or if first condition is satisfied the schedule will be created
no behaviour change from before

apps/sage_intacct/tasks.py
```diff
@@ -1235,7 +1237,7 @@ def schedule_sage_intacct_reimbursement_creation(sync_fyle_to_sage_intacct_payme
                 }
             )
 
-        if not sync_fyle_to_sage_intacct_payments:
+        if not sync_fyle_to_sage_intacct_payments or configuration.reimbursable_expenses_object != 'EXPENSE_REPORT':
```
Shwetabhk, line 1240: use and not or

Shwetabhk (approved):

Shwetabhk (approved):

PR #304, opened 2023-06-07
Refactor Dependent Fields Import

Shwetabhk:


apps/fyle/models.py
```diff
@@ -510,4 +510,25 @@ class ExpenseFilter(models.Model):
     updated_at = models.DateTimeField(auto_now=True, help_text='Updated at')
 
     class Meta:
-        db_table = 'expense_filters'
\ No newline at end of file
+        db_table = 'expense_filters'
+
+
+class DependentField(models.Model):
```
Shwetabhk, line 516:
```suggestion
class DependentFieldSetting(models.Model):
```

apps/fyle/models.py
```diff
@@ -510,4 +510,25 @@ class ExpenseFilter(models.Model):
...
+    cost_code_field_id = models.IntegerField(help_text='Fyle Cost Code Field ID')
+    cost_type_field_name = models.CharField(max_length=255, help_text='Fyle Cost Type Field Name')
+    cost_type_field_id = models.IntegerField(help_text='Fyle Cost Type Field ID')
+    workspace = models.OneToOneField(Workspace, on_delete=models.PROTECT, help_text='Reference to Workspace')
+    last_successful_import_at = models.DateTimeField(null=True, help_text='Last Successful Import At')
+    created_at = models.DateTimeField(auto_now_add=True, help_text='Created at')
+    updated_at = models.DateTimeField(auto_now=True, help_text='Updated at')
+
+    class Meta:
+        db_table = 'dependent_fields'
```
Shwetabhk, line 534:
```suggestion
        db_table = 'dependent_field_settings'
```

apps/fyle/serializers.py
```diff
@@ -72,4 +72,17 @@ def create(self, validated_data):
             defaults=validated_data
         )
 
-        return expense_filter
\ No newline at end of file
+        return expense_filter
+
+
+class DependentFieldSerializer(serializers.ModelSerializer):
```
Shwetabhk, line 78: Same

apps/fyle/urls.py
```diff
@@ -16,10 +16,11 @@
 from django.urls import path
 import itertools
 
-from .views import CustomFieldView, ExpenseFilterView, ExpenseGroupExpenseView, ExpenseGroupView, ExpenseGroupScheduleView, ExpenseGroupByIdView, \
-    ExpenseView, EmployeeView, CategoryView, ProjectView, CostCenterView, FyleFieldsView, \
-        ExpenseAttributesView, ExpenseGroupSettingsView, RefreshFyleDimensionView, SyncFyleDimensionView, \
-        ExpenseGroupCountView
+from .views import (
+    CustomFieldView, ExpenseFilterView, ExpenseGroupExpenseView, ExpenseGroupView, ExpenseGroupScheduleView, ExpenseGroupByIdView,
```
Shwetabhk, line 20: Cand we add line breaks and have just 3 in 1 line?

apps/fyle/urls.py
```diff
@@ -44,7 +45,8 @@
     path('fyle_fields/', FyleFieldsView.as_view()),
     path('expense_filters/', ExpenseFilterView.as_view(), name='expense-filters'),
     path('expenses/', ExpenseView.as_view(), name='expenses'),
-    path('custom_fields/', CustomFieldView.as_view(), name='custom-field')
+    path('custom_fields/', CustomFieldView.as_view(), name='custom-field'),
+    path('dependent_fields/', DependentFieldView.as_view(), name='dependent-field')
```
Shwetabhk, line 49:
```suggestion
    path('dependent_field_settings/', DependentFieldSettingView.as_view(), name='dependent-field')
```

apps/fyle/views.py
```diff
@@ -19,9 +19,11 @@
 
 from .tasks import create_expense_groups, schedule_expense_group_creation
 from .helpers import check_interval_and_sync_dimension, sync_dimensions
-from .models import Expense, ExpenseFilter, ExpenseGroup, ExpenseGroupSettings
-from .serializers import ExpenseFilterSerializer, ExpenseGroupExpenseSerializer, ExpenseGroupSerializer, ExpenseSerializer, ExpenseFieldSerializer, \
-    ExpenseGroupSettingsSerializer
+from .models import Expense, ExpenseFilter, ExpenseGroup, ExpenseGroupSettings, DependentField
+from .serializers import (
+    ExpenseFilterSerializer, ExpenseGroupExpenseSerializer, ExpenseGroupSerializer, ExpenseSerializer, ExpenseFieldSerializer,
```
Shwetabhk, line 24: Cand we add line breaks and have just 3 in 1 line?

apps/fyle/views.py
```diff
@@ -432,3 +434,11 @@ def get(self, request, *args, **kwargs):
             status=status.HTTP_200_OK
         )
 
+
+class DependentFieldView(generics.CreateAPIView, generics.RetrieveUpdateAPIView):
```
Shwetabhk, line 438:
```suggestion
class DependentFieldSettingView(generics.CreateAPIView, generics.RetrieveUpdateAPIView):
```

Shwetabhk: Are we using updates?
of not then just Create and Retrieve will do right?

apps/sage_intacct/dependent_fields.py
```diff
@@ -0,0 +1,130 @@
...
+logger = logging.getLogger(__name__)
+logger.level = logging.INFO
+
+
+def post_dependent_expense_field_values(workspace_id: int, dependent_field: DependentField, platform: PlatformConnector = None):
+    if not platform:
+        platform = connect_to_platform(workspace_id)
+
+    # JSONBAgg get both status
+    # projects = CostType.objects.filter(workspace_id=workspace_id).values('project_name').annotate(tasks=JSONBAgg('task_name', distinct=True))
```
Shwetabhk, line 28: Lets keep this for now, might be handy later when we try to do auto sync

apps/sage_intacct/dependent_fields.py
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    if dependent_field.last_successful_import_at:
+        filters['updated_at__gte'] = dependent_field.last_successful_import_at
+
+    projects = CostType.objects.filter(**filters).values('project_name').annotate(tasks=ArrayAgg('task_name', distinct=True))
+    tasks = CostType.objects.filter(**filters).values('task_name').annotate(cost_types=ArrayAgg('name', distinct=True))
+
+    # TODO: check auto-sync possibility, via sdk
+    for project in projects:
```
Shwetabhk, line 42: Were we not gonna do this in chain?
first cost code upload, then cost type upload?

apps/sage_intacct/dependent_fields.py
```diff
@@ -0,0 +1,130 @@
...
+                'parent_expense_field_id': dependent_field.project_field_id,
+                'parent_expense_field_value': project['project_name'],
+                'expense_field_id': dependent_field.cost_code_field_id,
+                'expense_field_value': task,
+                'is_enabled': True
+            } for task in project['tasks']
+        ]
+        platform.expense_fields.bulk_post_dependent_expense_field_values(payload)
+
+    for task in tasks:
```
Shwetabhk, line 54: Will there be any advantage of breaking these loops out in a generator function?

Shwetabhk: Think about it maybe

apps/sage_intacct/dependent_fields.py
```diff
@@ -0,0 +1,130 @@
...
+def import_dependent_fields_to_fyle(workspace_id: str):
+    dependent_field = DependentField.objects.get(workspace_id=workspace_id)
+    platform = connect_to_platform(workspace_id)
+
+    sync_sage_intacct_attributes('COST_TYPE', workspace_id)
+
+    try:
+        post_dependent_expense_field_values(workspace_id, dependent_field, platform)
+
+    except (SageIntacctCredential.DoesNotExist, InvalidTokenError):
```
Shwetabhk, line 79: later on when we rework imports, lets use decorators for exceptions here too

apps/sage_intacct/helpers.py
```diff
@@ -57,8 +65,7 @@ def sync_dimensions(si_credentials: SageIntacctCredential, workspace_id: int, di
         dimensions = [
             'locations', 'customers', 'departments', 'tax_details', 'projects', 
             'expense_payment_types', 'classes', 'charge_card_accounts','payment_accounts', 
-            'vendors', 'employees', 'accounts', 'expense_types', 'items', 'user_defined_dimensions',
-            'tasks', 'cost_types'
+            'vendors', 'employees', 'accounts', 'expense_types', 'items', 'user_defined_dimensions', 'cost_types'
```
Shwetabhk, line 68: Sync dimensions are called synchronously
cost_types will be synced in the API this way lets do cost_types in `async_task`
Nothing should happen in the API!

apps/sage_intacct/utils.py
```diff
@@ -261,62 +264,27 @@ def sync_payment_accounts(self):
...
         Sync of Sage Intacct Cost Types
         """
+        args = {
+            'field': 'STATUS',
+            'value': 'active'
+        }
 
-        cost_types = self.connection.cost_types.get_all()
-        cost_types_attributes = []
+        latest_synced_cost_type = CostType.objects.filter(workspace_id=self.workspace_id).order_by('-updated_at').first()
```
Shwetabhk, line 276: we're storing last_imported_at in dependent fields table right? Why not just use that?

apps/fyle/views.py
ashwin1111: we're using patch to turn off schedule (turn off toggle in UI)

apps/sage_intacct/dependent_fields.py
ashwin1111: task upload is dependent on cost type. if we're using chains, there is no way to stop 2nd chain of 1st chain stops. Overall this block will be run by the worker, so things are async in nature.

apps/sage_intacct/dependent_fields.py
ashwin1111: not much addon other than readability, for now, I'll move these 2 post to a separate functions to keep the function smaller

apps/sage_intacct/dependent_fields.py
ashwin1111: sure, right now we don't have that in Intacct. Along with import rework we can do that

apps/sage_intacct/utils.py
ashwin1111: incase if there are failed runs, we might end up getting old ones again and again, but kind of an edge case, will use the existing timestamp from deps table

ashwin1111 (approved):

NileshPant1999 (approved):

PR #306, opened 2023-06-15
feature: advance import settings apis

NileshPant1999:
GET and PUT http://localhost:8004/api/v2/workspaces/1/advanced_import_settings

```json
{
    "configurations": {
        "change_accounting_period": true,
        "sync_fyle_to_sage_intacct_payments": true,
        "sync_sage_intacct_to_fyle_payments": false,
        "auto_create_destination_entity": false,
        "memo_structure": [
            "merchant",
            "purpose"
        ]
    },
    "general_mappings": {
        "default_location": {
            "name": "General Construction - San Antonio",
            "id": "GC-SAN"
        },
        "payment_account": {
            "name": "First Community Bank",
            "id": "100"
        },
        "default_department": {
            "name": "",
            "id": ""
        },
        "default_class": {
            "name": "",
            "id": ""
        },
        "default_project": {
            "name": "",
            "id": ""
        },
        "default_item": {
            "name": "Mechanical",
            "id": "CN015"
        },
        "use_intacct_employee_departments": false,
        "use_intacct_employee_locations": false
    },
    "workspace_schedules": {
        "enabled": true,
        "interval_hours": 24,
        "additional_email_options": [],
        "emails_selected": [
            "fyle@fyle.in"
        ]
    },
    "workspace_id": 1
}
```


apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -0,0 +1,201 @@
...
+        if not expense_group_settings['corporate_credit_card_expense_group_fields']:
+            expense_group_settings['corporate_credit_card_expense_group_fields'] = ['employee_email', 'report_id', 'fund_source']
+
+        if not expense_group_settings['reimbursable_export_date_type']:
+            expense_group_settings['reimbursable_export_date_type'] = 'current_date'
+
+        if not expense_group_settings['ccc_export_date_type']:
+            expense_group_settings['ccc_export_date_type'] = 'current_date'
+
+        expense_group_settings['import_card_credits'] = False
```
ashwin1111, line 167: why are we hard coding this?

apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -0,0 +1,201 @@
...
+            defaults={
+                'auto_map_employees': configurations['auto_map_employees'],
+                'reimbursable_expenses_object': configurations['reimbursable_expenses_object'], 
+                'corporate_credit_card_expenses_object': configurations['corporate_credit_card_expenses_object'],
+                'employee_field_mapping': configurations['employee_field_mapping']
+            }
+        )
+
+        if not expense_group_settings['reimbursable_expense_group_fields']:
+            expense_group_settings['reimbursable_expense_group_fields'] = ['employee_email', 'report_id', 'fund_source']
```
ashwin1111, line 156: let's add claim number too, it will cause issues around memo

apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -0,0 +1,201 @@
...
+
+        if not expense_group_settings['corporate_credit_card_expense_group_fields']:
+            expense_group_settings['corporate_credit_card_expense_group_fields'] = ['employee_email', 'report_id', 'fund_source']
```
ashwin1111, line 159: let's add claim number too, it will cause issues around memo

apps/workspaces/apis/import_settings/serializers.py
```diff
@@ -0,0 +1,172 @@
...
+            self.field_name: data
+        }
+
+
+class MappingSettingSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = MappingSetting
+        list_serializer_class = MappingSettingFilteredListSerializer
+        fields = [
+            'source_field',
```
ashwin1111, line 55: this is not required

apps/workspaces/apis/import_settings/serializers.py
```diff
@@ -0,0 +1,172 @@
...
+                    'import_to_fyle': True,
+                    'is_custom': False
+                })
+        
+            for setting in mapping_settings:
+                MappingSetting.objects.update_or_create(
+                    destination_field=setting['destination_field'],
+                    workspace_id=instance.id,
+                    defaults={
+                        'source_field': setting['source_field'],
```
ashwin1111, line 147: not needed

apps/workspaces/apis/import_settings/triggers.py
```diff
@@ -0,0 +1,46 @@
...
+        for setting in self.__mapping_settings:
+            if setting['destination_field'] not in destination_fields:
+                destination_fields.append(setting['destination_field'])
+        
+        #Now that we have all the system necessary mapping-settings and the mapping-settings in the payload
+        #This query will take care of deleting all the redundant mapping-settings that are not required.
+        MappingSetting.objects.filter(
+            ~Q(destination_field__in=destination_fields),
+            workspace_id=self.__workspace_id
+        ).delete()
```
ashwin1111, line 44: this behaviour might change in new UI, will confirm after checking design

apps/workspaces/models.py
```diff
@@ -10,6 +10,19 @@
 User = get_user_model()
 
 
+ONBOARDING_STATE_CHOICES = (
+    ('CONNECTION', 'CONNECTION'),
```
ashwin1111, line 14: can we also have 1 state for location entity? right now there is no distinguish.

apps/workspaces/models.py
```diff
@@ -23,6 +36,10 @@ class Workspace(models.Model):
     ccc_last_synced_at = models.DateTimeField(help_text='Datetime when ccc expenses were pulled last', null=True)
     source_synced_at = models.DateTimeField(help_text='Datetime when source dimensions were pulled', null=True)
     destination_synced_at = models.DateTimeField(help_text='Datetime when destination dimensions were pulled', null=True)
+    onboarding_state = models.CharField(
+        max_length=50, choices=ONBOARDING_STATE_CHOICES, default=get_default_onboarding_state,
+        help_text='Onboarding status of the workspace', null=True
+    )
```
ashwin1111, line 42: add script to fill data

apps/workspaces/tasks.py
```diff
@@ -212,14 +212,3 @@ def async_update_fyle_credentials(fyle_org_id: str, refresh_token: str):
...
-def delete_cards_mapping_settings(configuration: Configuration):
-    if configuration.corporate_credit_card_expenses_object != 'CHARGE_CARD_TRANSACTION' :
-        mapping_setting = MappingSetting.objects.filter(
-            workspace_id=configuration.workspace_id,
-            source_field='CORPORATE_CARD',
-            destination_field='CHARGE_CARD_NUMBER'
-        ).first()
-
-        if mapping_setting:
-            mapping_setting.delete()
```
ashwin1111, line 225: check if old api is not affected

apps/workspaces/apis/export_settings/serializers.py
NileshPant1999: I am fixing this

apps/workspaces/apis/advanced_settings/triggers.py
```diff
@@ -0,0 +1,14 @@
+from apps.workspaces.models import Configuration
+from apps.sage_intacct.helpers import schedule_payment_sync
+
+
+class AdvancedConfigurationsTriggers:
+    """
+    Class containing all triggers for advanced_configurations
+    """
+    @staticmethod
+    def run_configurations_triggers(configurations_instance: Configuration):
```
Shwetabhk, line 10: Function name should reflect if it is pre_save or post_save

apps/workspaces/apis/advanced_settings/triggers.py
```diff
@@ -0,0 +1,14 @@
...
+        """
+        Run workspace general settings triggers
+        """
+        schedule_payment_sync(configurations_instance)
```
Shwetabhk, line 14: Can you confirm if this is not already called in the Signals? If it is then we don't need this whole triggers file, we don't wanna call it twice, only the things which might not be handled in Signals should be there in the triggers

apps/workspaces/apis/advanced_settings/serializers.py
```diff
@@ -0,0 +1,186 @@
...
+
+        GeneralMapping.objects.update_or_create(
+            workspace=instance,
+            defaults={
+                'payment_account_name': general_mappings.get('payment_account').get('name'),
+                'payment_account_id': general_mappings.get('payment_account').get('id')
+            }
+        )
+
+        schedule_sync(
```
Shwetabhk, line 160: This should be in the trigger or a signal and not here

Shwetabhk (approved):

apps/workspaces/apis/advanced_settings/triggers.py
NileshPant1999: this is there in signals @Shwetabhk, i am removing this trigger file 👍🏻

apps/workspaces/apis/advanced_settings/serializers.py
NileshPant1999: we are doing this way only in qbo-api that is why added it here, moving it to a trigger function

PR #318, opened 2023-06-30
disable the expense attribute which are renamed in intacct

Ashutosh619-sudo:


apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
 
         Mapping.bulk_create_mappings(paginated_si_attributes, 'PROJECT', destination_field, workspace_id)
     
+    expense_attributes_count = ExpenseAttribute.objects.filter(attribute_type='PROJECT',workspace_id=workspace_id).count()
+    expense_attribute_to_be_changed = []
```
ashwin1111, line 167: expense_attribute_to_be_disabled

apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
 
         Mapping.bulk_create_mappings(paginated_si_attributes, 'PROJECT', destination_field, workspace_id)
     
+    expense_attributes_count = ExpenseAttribute.objects.filter(attribute_type='PROJECT',workspace_id=workspace_id).count()
```
ashwin1111, line 166: add auto_mapped in filter

apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
...
+    for offset in range(0, expense_attributes_count, page_size):
+        limit = offset + page_size
+
+        fyle_projects = ExpenseAttribute.objects.filter(
+            workspace_id=workspace_id,
+            attribute_type="PROJECT",
+            auto_mapped=True
+        ).order_by('value', 'id')[offset:limit]
+
+        expense_field_values = list(
```
ashwin1111, line 177: project_names

apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
 
         Mapping.bulk_create_mappings(paginated_si_attributes, 'PROJECT', destination_field, workspace_id)
```
ashwin1111, line 164: move all new code to a separate function, also call this function only when destiation_field = 'PROJECT'

apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
...
+            set(field.value for field in fyle_projects)
+        )
+
+        destination_field_values = DestinationAttribute.objects.filter(
+        attribute_type=destination_field,
+        workspace_id=workspace_id,
+        value__in=expense_field_values
+        ).values_list('value', flat=True)
+
+        for field in fyle_projects:
```
ashwin1111, line 187: call it fyle_project instead of field

apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
...
+            workspace_id=workspace_id,
+            attribute_type="PROJECT",
+            auto_mapped=True
+        ).order_by('value', 'id')[offset:limit]
+
+        expense_field_values = list(
+            set(field.value for field in fyle_projects)
+        )
+
+        destination_field_values = DestinationAttribute.objects.filter(
```
ashwin1111, line 181: call it netsuite_projects

apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
...
+        ).values_list('value', flat=True)
+
+        for field in fyle_projects:
+            if field.value not in destination_field_values:
+                field.active = False
+                field.save()
+                expense_attribute_to_be_changed.append(field)
+        
+        fyle_payload = create_fyle_projects_payload(projects=[],existing_project_names=[],updated_projects=expense_attribute_to_be_changed)
+        platform.projects.post_bulk(fyle_payload)
```
ashwin1111, line 194: post only when fyle_payload exist

apps/mappings/tasks.py
```diff
@@ -163,6 +163,37 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
...
+        platform.projects.sync()
```
ashwin1111, line 195: we're doing similar operations in the below block too. append both arrays and just post once and sync once. Just add a comment that the new code solves the rename problem in accounting software

ashwin1111 (approved):

PR #321, opened 2023-07-05
sage intacct api redesign errors module

NileshPant1999:


apps/sage_intacct/tasks.py
```diff
@@ -398,6 +398,17 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
     if not errors:
         errors.append(exception.response)
 
+    Error.objects.update_or_create(
+            workspace_id=expense_group.workspace_id,
+            expense_group=expense_group,
+            defaults={
+                'type': 'INTACCT_ERROR',
+                'error_title': 'this is a error',
+                'error_detail': 'error dedsf',
```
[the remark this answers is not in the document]
ashwin1111: copy from DB

apps/mappings/signals.py
```diff
@@ -20,11 +20,45 @@
...
 logger = logging.getLogger(__name__)
 
 
+@receiver(post_save, sender=Mapping)
+def resolve_post_mapping_errors(sender, instance: Mapping, **kwargs):
+    """
+    Resolve errors after mapping is created
+    """
+    if instance.source_type in ('TAX_GROUP'):
+        error = Error.objects.filter(expense_attribute_id=instance.source_id).first()
```
ashwin1111, line 34: .filter(...).update(...)

apps/mappings/signals.py
```diff
@@ -20,11 +20,45 @@
...
+            error.is_resolved = True
+            error.save()
+            
+
+@receiver(post_save, sender=CategoryMapping)
+def resolve_post_category_mapping_errors(sender, instance: Mapping, **kwargs):
+    """
+    Resolve errors after mapping is created
+    """
+    error = Error.objects.filter(expense_attribute_id=instance.source_category_id).first()
```
ashwin1111, line 45: same as above

apps/mappings/signals.py
```diff
@@ -20,11 +20,45 @@
...
+        error.is_resolved = True
+        error.save()
+
+
+@receiver(post_save, sender=EmployeeMapping)
+def resolve_post_employees_mapping_errors(sender, instance: Mapping, **kwargs):
+    """
+    Resolve errors after mapping is created
+    """
+    error = Error.objects.filter(expense_attribute_id=instance.source_employee_id).first()
```
ashwin1111, line 56: same as above

apps/mappings/tasks.py
```diff
@@ -13,22 +13,43 @@
...
+
+    if errored_attribute_ids:
+        mapped_attribute_ids = []
+
+        mapped_attribute_ids: List[int] = Mapping.objects.filter(
+            source_id__in=errored_attribute_ids
+        ).values_list('source_id', flat=True)
+
+        if mapped_attribute_ids:
+            Error.objects.filter(expense_attribute_id__in=mapped_attribute_ids).update(is_resolved=True)
```
ashwin1111, line 50: employee and category mappings are not handled

apps/sage_intacct/tasks.py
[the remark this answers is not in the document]
ashwin1111: sorry, I thought this is from tests. @NileshPant1999 can you get a dump from DB and analyse pattern?

apps/sage_intacct/tasks.py
```diff
@@ -420,6 +431,13 @@ def __validate_expense_group(expense_group: ExpenseGroup, configuration: Configu
             'message': 'General mappings not found'
         })
 
+    
+    employee_attribute = ExpenseAttribute.objects.filter(
+                value=expense_group.description.get('employee_email'),
+                workspace_id=expense_group.workspace_id,
+                attribute_type='EMPLOYEE'
+    ).first()
```
ashwin1111, line 439: indent is off

apps/workspaces/apis/errors/views.py
```diff
@@ -0,0 +1,32 @@
...
+            'workspace__id': self.kwargs.get('workspace_id')
+        }
+
+        if is_resolved and is_resolved.lower() == 'true':
+            is_resolved = True
+        elif is_resolved and is_resolved.lower() == 'false':
+            is_resolved = False
+
+        if is_resolved is not None:
+            params['is_resolved'] = is_resolved
```
ashwin1111, line 27: simplify this, lot of gymnastics here

apps/mappings/tasks.py
```diff
@@ -42,12 +42,47 @@ def resolve_expense_attribute_errors(
...
+                }
+            else:
+                params = {
+                    'source_category_id__in': errored_attribute_ids,
+                    'categorymapping__destination_account_id__isnull': False
+                }
+                
+            mapped_attribute_ids: List[int] = CategoryMapping.objects.filter(
+                **params
+            ).values_list('source_employee_id', flat=True)
```
ashwin1111: move whole mapped_attribute_ids to a separate function, func got bigger

apps/workspaces/apis/errors/views.py
```diff
@@ -11,22 +11,17 @@ class ErrorsView(generics.ListAPIView):
...
 
-        if is_resolved and is_resolved.lower() == 'true':
-            is_resolved = True
-        elif is_resolved and is_resolved.lower() == 'false':
-            is_resolved = False
+        if type:
+            params['type'] = type
 
         if is_resolved is not None:
+            is_resolved = is_resolved.lower() == 'true'
```
ashwin1111, line 24: from the client, we are only using is_resolved for false, can just convert 'false' to False. There is no usecase of is resolved = true

apps/sage_intacct/tasks.py
```diff
@@ -403,8 +403,8 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
             expense_group=expense_group,
             defaults={
                 'type': 'INTACCT_ERROR',
-                'error_title': 'this is a error',
-                'error_detail': 'error dedsf',
+                'error_title': 'Sage Intacct Internal Error',
```
ashwin1111, line 406: we shouldn't hard code title

apps/sage_intacct/tasks.py
```diff
@@ -403,8 +403,8 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
...
+                'error_detail': errors[0],
```
ashwin1111, line 407: did we check data in DB? only the 1st element helps in every case?

NileshPant1999: yes @ashwin1111 i analysed the pattern, the first element contains all the major information

ashwin1111: can you add it to some doc / sheet, would like to check it out.

apps/sage_intacct/tasks.py
```diff
@@ -395,16 +400,23 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
...
+            error_title = errors[0]['correction'] if ('correction' in errors[0] and errors[0]['correction']) else errors[0]['short_description']
+            error_msg = errors[0]['long_description']
+
     if not errors:
         errors.append(exception.response)
+        error_title = 'Failed to create {0} in your Sage Intacct account.'.format(export_type)
+        error_msg = exception.response
 
+    print('ello', errors)
+    print(error_msg, error_title)
```
ashwin1111, line 412: Screenshot 2023-07-14 at 4 37 31 PM

apps/sage_intacct/tasks.py
```diff
@@ -383,6 +385,9 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
                          if ('correction' in error and error['correction']) else 'Not available'
                 })
 
+            error_title = errors[0]['correction'] if ('correction' in errors[0] and errors[0]['correction']) else errors[0]['short_description']
+            error_msg = errors[0]['long_description']
+
```
ashwin1111, line 390: duplicate logic, can just move 403-404 1 block to the left

apps/sage_intacct/tasks.py
```diff
@@ -395,16 +400,23 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
                     if ('correction' in error and error['correction']) else 'Not available'
             })
 
+            error_title = errors[0]['correction'] if ('correction' in errors[0] and errors[0]['correction']) else errors[0]['short_description']
```
ashwin1111, line 403: 'correction' key will always be present, no need of checking this

ashwin1111: this is also not taken care
Screenshot 2023-07-14 at 4 45 34 PM

apps/sage_intacct/tasks.py
```diff
@@ -395,16 +400,23 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
...
+            error_msg = errors[0]['long_description']
+
     if not errors:
         errors.append(exception.response)
+        error_title = 'Failed to create {0} in your Sage Intacct account.'.format(export_type)
+        error_msg = exception.response
```
ashwin1111, line 409: this won't be a strong, add some common string like 'Something unexpected has happened'. Also, can do these directly in line 368, 369

apps/sage_intacct/tasks.py
```diff
@@ -365,8 +365,8 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
     logger.info(exception.response)
     
     errors = []
-    error_title = None
-    error_msg = None
+    error_title = 'Failed to create {0} in your Sage Intacct account.'.format(export_type)
+    error_msg = 'Something unexpected happened with workspace_id: %s %s'.format(task_log.workspace_id, exception.response)
```
ashwin1111, line 369: workspace id is not needed, this is exposed in UI

apps/sage_intacct/tasks.py
```diff
@@ -400,16 +397,13 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
                     if ('correction' in error and error['correction']) else 'Not available'
             })
 
-            error_title = errors[0]['correction'] if ('correction' in errors[0] and errors[0]['correction']) else errors[0]['short_description']
-            error_msg = errors[0]['long_description']
+        error_title = errors[0]['correction'] if (errors[0]['correction'] != 'Not available') else errors[0]['short_description']
```
ashwin1111, line 402: if errors[0]['correction']  and != 'not available'

ashwin1111 (approved):

apps/sage_intacct/tasks.py
NileshPant1999: @ashwin1111 if this is exposed in UI we should also not add the exception right ??

ashwin1111: yes we shouldn't, we won't land in this case, just a fallback

PR #322, opened 2023-07-05
fyle and expense group redesign

NileshPant1999:


apps/workspaces/views.py
```diff
@@ -489,29 +489,19 @@ def get(self, request, *args, **kwargs):
...
 
-class LastExportDetailView(viewsets.ViewSet):
+class LastExportDetailView(generics.RetrieveAPIView):
     """
     Last Export Details
     """
-    serializer_class = LastExportDetailSerializer
+    
+    authentication_classes = []
+    permission_classes = []
```
ashwin1111, line 498: remove these

scripts/sql/scripts/022_fill_employee_name_in_expenses_and_expense_groups.sql
```diff
@@ -18,7 +18,7 @@ set
 from 
   ws 
 where 
-  workspace_id = ws.expense_attributes_workspace_id;
+  employee_email = ws.expense_attributes_employee_email;
```
ashwin1111, line 21: which script fills in employee_email to expense groups? can you point me to that?

apps/workspaces/views.py
NileshPant1999: ohh yes, was testing api will remove these

scripts/sql/scripts/022_fill_employee_name_in_expenses_and_expense_groups.sql
NileshPant1999: @ashwin1111 we are already storing employee email in the expense group table no in description column.

ashwin1111 (approved):

PR #323, opened 2023-07-10
task filter for polling

NileshPant1999:


apps/tasks/views.py
```diff
@@ -20,15 +20,24 @@ def get_queryset(self):
         """
         Return task logs in workspace
         """
-        task_status = self.request.query_params.getlist('status')
-        task_type = ['CREATING_AP_PAYMENT', 'CREATING_REIMBURSEMENT']
+        task_type = self.request.query_params.get('task_type')
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+        task_status = self.request.query_params.get('status')
```
NileshPant1999, line 26: @ashwin1111 this will need some app changes, if you want me to write this in a way it will not break current frontend please let me know otherwise i will fix this in app

ashwin1111 (approved):

PR #328, opened 2023-07-18
NoPrivilegeError exception handled

ruuushhh:


apps/fyle/tasks.py
```diff
@@ -7,6 +7,7 @@
 from django_q.tasks import async_task
 
 from fyle_integrations_platform_connector import PlatformConnector
+from sageintacctsdk.exceptions import NoPrivilegeError
```
ashwin1111, line 10: this should from Fyle end, we call only platform apis here

apps/fyle/tasks.py
```diff
@@ -145,6 +146,14 @@ def create_expense_groups(workspace_id: int, fund_source: List[str], task_log: T
             task_log.status = 'COMPLETE'
             task_log.save()
 
+    except NoPrivilegeError:
+        logger.info('Insufficient permission to access the requested module')
```
ashwin1111, line 150: change logger and detail to Invalid Fyle Credentials / Admin is disabled

ashwin1111 (approved):

PR #329, opened 2023-07-19
Import rewrite

labhvam5:


apps/mappings/imports/modules/base.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def update_import_log_post_import(self, is_last_batch: bool, import_log: ImportLog):
+        """
+        Update Import Log Post Import
+        :param is_last_batch: bool
+        :param import_log: ImportLog object
+        """
+        if is_last_batch:
+            import_log.last_successful_run_at = datetime.now()
+            import_log.processed_batches_count += 1
```
ashwin1111, line 308: where are we resetting this value to 0? I am not able to locate this code

apps/mappings/imports/modules/projects.py
```diff
@@ -0,0 +1,54 @@
...
+from apps.mappings.imports.modules.base import Base
+from fyle_accounting_mappings.models import DestinationAttribute
+
+
+class Project(Base):
+    """
+    Class for Projects module
+    """
+    def __init__(self, workspace_id: int, destination_field: str, sync_after: datetime):
+        super().__init__(workspace_id, 'PROJECT' , destination_field, 'projects', sync_after, batch_size=2)
```
ashwin1111, line 12: `super().__init__(workspace_id=workspace_id, source_field='PROJECT' , destination_field, 'projects', sync_after, batch_size=2)`
for better readability, do it for all args, also change batch size

apps/mappings/imports/schedules.py
```diff
@@ -0,0 +1,44 @@
+from datetime import datetime
+from django_q.models import Schedule
+from apps.workspaces.models import Configuration
+from fyle_accounting_mappings.models import MappingSetting
+
+
+def schedule_or_delete_fyle_import_tasks(configuration: Configuration, instance: MappingSetting = None):
```
ashwin1111, line 7: instance will also be sent no, why default it?

apps/mappings/imports/schedules.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Schedule or delete Fyle import tasks based on the configuration.
+    :param configuration: Workspace Configuration Instance
+    :param instance: Mapping Setting Instance
+    :return: None
+    """
+    # Check if there is a task to be scheduled
+    if instance:
```
ashwin1111, line 15: what is the case when instance would be null

apps/mappings/imports/schedules.py
```diff
@@ -0,0 +1,44 @@
...
+            workspace_id=configuration.workspace_id,
+            source_field=instance.source_field
+        ).first()
+
+    # Check if any of the configuration flags are True
+    if (instance and task_to_be_scheduled) or (
+        configuration.import_categories or
+        configuration.import_projects or
+        configuration.import_vendors_as_merchants or
+        configuration.import_tax_codes
```
ashwin1111, line 27: since we are just releasing this for projects, why bring in all other apis?

apps/mappings/imports/modules/base.py
```diff
@@ -0,0 +1,314 @@
...
+    The Base class for all the modules
+    """
+    def __init__(
+            self,
+            workspace_id: int,
+            source_field: str,
+            destination_field: str,
+            class_name: str,
+            sync_after:datetime,
+            batch_size: int = 200
```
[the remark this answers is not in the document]
labhvam5: I think we should remove this.

apps/mappings/imports/modules/base.py
```diff
@@ -0,0 +1,314 @@
...
+        self.sync_expense_attributes(platform)
+
+        self.create_mappings()
+
+    def create_mappings(self):
+        """
+        Create mappings
+        """
+        paginated_destination_attributes_without_duplicates = []
+        paginated_destination_attributes = DestinationAttribute.objects.filter(
```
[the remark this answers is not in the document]
labhvam5: Have fixed this

apps/mappings/imports/schedules.py
labhvam5: This `schedule_or_delete_fyle_import_tasks()` is also called from the `run_post_configration_triggers()` -> `schedule_or_delete_auto_mapping_tasks()`. 

In this case we only get the configuration and instance will not be sent. That is why we have defaulted it to `None` .

apps/mappings/imports/modules/base.py
```diff
@@ -0,0 +1,314 @@
...
+class Base:
+    """
+    The Base class for all the modules
+    """
+    def __init__(
+            self,
+            workspace_id: int,
+            source_field: str,
+            destination_field: str,
+            class_name: str,
```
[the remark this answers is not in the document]
labhvam5: have gone ahead with `platform_class_name`

apps/mappings/imports/modules/base.py
labhvam5: in the `check_import_log_and_start_import()` when we start the import we set the count back to 0.

apps/mappings/imports/schedules.py
labhvam5: we wont be adding the schedule .

apps/mappings/imports/schedules.py
labhvam5: have removed the other condition.

apps/mappings/tasks.py
```diff
@@ -1291,11 +1141,8 @@ def auto_import_and_map_fyle_fields(workspace_id):
     if configuration.import_categories:
         chain.append('apps.mappings.tasks.auto_create_category_mappings', workspace_id)
 
-    if project_mapping and project_mapping.import_to_fyle:
-        chain.append('apps.mappings.tasks.auto_create_project_mappings', workspace_id)
-
-        if dependent_fields and dependent_fields.is_import_enabled:
-            chain.append('apps.sage_intacct.dependent_fields.import_dependent_fields_to_fyle', workspace_id)
+    if dependent_fields and dependent_fields.is_import_enabled:
```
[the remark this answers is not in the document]
labhvam5: Can you explain this point not able to understand this ?

apps/mappings/imports/modules/base.py
ashwin1111: got it

apps/mappings/imports/schedules.py
ashwin1111: not able to follow

apps/mappings/tasks.py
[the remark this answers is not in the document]
ashwin1111: project mapping condition is also needed, only if it is present we should call dependent fields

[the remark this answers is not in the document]
labhvam5: ohh okay .

apps/mappings/imports/schedules.py
labhvam5: If the instance is null then we only check if `configuration.import_projects` is enabled then in that case we add the schedule else we dont add the schedule and delete the existing schedule .

apps/mappings/tasks.py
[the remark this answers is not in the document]
labhvam5: have updated the condition to 
`if (project_mapping and project_mapping.import_to_fyle) and (dependent_fields and dependent_fields.is_import_enabled):`

ashwin1111 (approved):

PR #340, opened 2023-08-23
bug fix auto-create-category

labhvam5:


scripts/python/create-update-new-import-schedules.py
```diff
@@ -15,42 +12,49 @@
...
+        schedule_count = Schedule.objects.filter(func='apps.mappings.imports.queues.chain_import_fields_to_fyle').count()
+        import_task_workspace_ids = Configuration.objects.filter(import_categories=True, import_vendors_as_merchants=True).values_list('workspace_id', flat=True)
+        dependent_field_workspace_ids = DependentFieldSetting.objects.filter(is_import_enabled=True).values_list('workspace_id', flat=True)
+        combined_workspace_ids = []
+        combined_workspace_ids.extend(import_task_workspace_ids)
+        combined_workspace_ids.extend(dependent_field_workspace_ids)
+        combined_workspace_ids = list(set(combined_workspace_ids))
+        combined_workspace_ids_count = len(combined_workspace_ids)
+        auto_import_schedule_count = Schedule.objects.filter(func='apps.mappings.tasks.auto_import_and_map_fyle_fields').count()
+
```
ashwin1111, line 50: there shouldn't be any empty lines inside loop since we're running it in shell

ashwin1111 (approved):

PR #342, opened 2023-08-24
dependent field bug fix

labhvam5:


apps/mappings/helpers.py
```diff
@@ -27,7 +29,10 @@ def schedule_or_delete_fyle_import_tasks(configuration: Configuration):
     :param configuration: Workspace Configuration Instance
     :return: None
     """
-    if configuration.import_categories or configuration.import_vendors_as_merchants:
+    project_mapping = MappingSetting.objects.filter(source_field='PROJECT', workspace_id=configuration.workspace_id).first()
+    dependent_fields = DependentFieldSetting.objects.filter(workspace_id=configuration.workspace_id).first()
+
+    if configuration.import_categories or configuration.import_vendors_as_merchants or (project_mapping and project_mapping.import_to_fyle and dependent_fields and dependent_fields.is_import_enabled):
```
ashwin1111, line 35: instead of doing `project_mapping and project_mapping.import_to_fyle`, add `import_to_fyle=True` to the ORM query, same for dependent_fields and dependent_fields.is_import_enabled

ashwin1111 (approved):

apps/mappings/helpers.py
labhvam5: did this for the `dependent_fields` as well .

PR #392, opened 2023-11-14
Logger settings updated

ruuushhh:


apps/sage_intacct/utils.py
```diff
@@ -26,6 +26,7 @@
 )
 
 logger = logging.getLogger(__name__)
+logger.level = logging.INFO
```
Shwetabhk, line 29: remove this and test without this

Shwetabhk (changes requested):

Shwetabhk (approved):

NileshPant1999 (approved):

PR #413, opened 2024-01-29
Export Merchant as Journal Entry Line Item

Ashutosh619-sudo:


apps/sage_intacct/models.py
```diff
@@ -920,6 +920,10 @@ def create_journal_entry_lineitems(expense_group: ExpenseGroup, configuration: C
                 default_employee_location_id is None else None
             employee_id = entity.destination_employee.destination_id if employee_mapping_setting == 'EMPLOYEE' else None
             vendor_id = entity.destination_vendor.destination_id if employee_mapping_setting == 'VENDOR' else None
+            if lineitem.fund_source=='CCC' and configuration.use_merchant_in_journal_line:
```
ashwin1111, line 923: lineitem.fund_source == 'CCC', once space before and after ==

apps/sage_intacct/models.py
```diff
@@ -920,6 +920,10 @@ def create_journal_entry_lineitems(expense_group: ExpenseGroup, configuration: C
...
+                vendor = DestinationAttribute.objects.filter(attribute_type='VENDOR', value=lineitem.vendor).first()
```
ashwin1111, line 924: value__iexact. Also add workspace id filter

apps/sage_intacct/models.py
ashwin1111: existing condition + `lineitem.vendor` - check if vendor is filled in expenses

ashwin1111 (approved):

PR #415, opened 2024-01-30
Post to integrations settings trigger

ruuushhh:


apps/workspaces/tasks.py
```diff
@@ -271,6 +273,26 @@ def async_update_fyle_credentials(fyle_org_id: str, refresh_token: str):
...
 
+def post_to_integration_settings(workspace_id: int, active: bool):
+    """
+    Post to integration settings
+    """
+    refresh_token = FyleCredential.objects.get(workspace_id=workspace_id).refresh_token
+    url = '{}/integrations/'.format(settings.INTEGRATIONS_SETTINGS_API)
+    payload = {
+        'tpa_id': settings.FYLE_CLIENT_ID,
+        'tpa_name': 'Fyle Intacct Integration',
```
ashwin1111, line 284: `Fyle Sage Intacct Integration`

fyle_intacct_api/settings.py
```diff
@@ -279,6 +279,7 @@
 FYLE_JOBS_URL = os.environ.get('FYLE_JOBS_URL')
 FYLE_APP_URL = os.environ.get('APP_URL')
 FYLE_EXPENSE_URL = os.environ.get('FYLE_APP_URL')
+INTEGRATIONS_SETTINGS_API = os.environ.get('INTEGRATIONS_SETTINGS_API')
```
ashwin1111, line 282: please add it to integrations central as well as create PR for fyle-deploy

scripts/python/post-integrations.py
```diff
@@ -0,0 +1,8 @@
+from apps.workspaces.tasks import post_to_integration_settings
+from apps.workspaces.models import Workspace
+
+workspaces = Workspace.objects.filter(onboarding_state='COMPLETE')
+
+for workspace in workspaces:
```
ashwin1111, line 6: do try catch inside for loop, many tokens would be dead

ashwin1111 (approved):

PR #420, opened 2024-01-30
Update Failed expenses in Fyle

ruuushhh:


apps/sage_intacct/tasks.py
```diff
@@ -911,6 +919,7 @@ def create_charge_card_transaction(expense_group: ExpenseGroup, task_log_id: int
         task_log.sage_intacct_errors = None
 
         task_log.save()
+        update_failed_expenses(expense_group.expenses.all(), True)
```
ashwin1111, line 922: we should update even incase of general exception

ashwin1111 (approved):

PR #427, opened 2024-02-06
Use legacy API for expense report

ashwin1111:


apps/sage_intacct/utils.py
```diff
@@ -849,8 +849,8 @@ def __construct_expense_report(self, expense_report: ExpenseReport,
             'description': expense_report.memo,
             'basecurr': expense_report.currency,
             'currency': expense_report.currency,
-            'eexpensesitems': {
-                'eexpensesitem': expsense_payload
+            'expenses': {
+                'expense': expsense_payload
```
Shwetabhk, line 853: can we fix the typo in expense_payload

Shwetabhk (approved):

PR #438, opened 2024-02-16
Added early payment sync to sage-intacct

Hrishabh17:
Please review.

### DIFF
--- a/apps/sage_intacct/tasks.py
+++ b/apps/sage_intacct/tasks.py
@@ -734,6 +734,8 @@ def create_expense_report(expense_group: ExpenseGroup, task_log_id: int, last_ex
         logger.exception('Something unexpected happened workspace_id: %s %s', task_log.workspace_id, task_log.detail)
 
     if last_export:
+        logger.info("| Last Export for Expense Report Creation | Content: {{WORKSPACE_ID : {} Configuration Expense Type: {}}}".format(expense_group.workspace_id, configuration.reimbursable_expenses_object))
+        create_sage_intacct_reimbursement(workspace_id=expense_group.workspace.id)
         update_last_export_details(expense_group.workspace_id)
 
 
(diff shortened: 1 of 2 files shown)


apps/sage_intacct/tasks.py
```diff
@@ -734,6 +734,8 @@ def create_expense_report(expense_group: ExpenseGroup, task_log_id: int, last_ex
         logger.exception('Something unexpected happened workspace_id: %s %s', task_log.workspace_id, task_log.detail)
 
     if last_export:
+        logger.info("| Last Export for Expense Report Creation | Content: {{WORKSPACE_ID : {} Configuration Expense Type: {}}}".format(expense_group.workspace_id, configuration.reimbursable_expenses_object))
+        create_sage_intacct_reimbursement(workspace_id=expense_group.workspace.id)
```
ashwin1111, line 738: let's call this func after update_last_export_details(), reason is payments sync is slightly a time consuming task, it shouldn't block update of last export detail

apps/sage_intacct/tasks.py
```diff
@@ -844,6 +846,8 @@ def create_bill(expense_group: ExpenseGroup, task_log_id: int, last_export: bool
         logger.exception('Something unexpected happened workspace_id: %s %s', task_log.workspace_id, task_log.detail)
     
     if last_export:
+        logger.info("| Last Export for Bill Creation | Content: {{WORKSPACE_ID : {} Configuration Expense Type: {}}}".format(expense_group.workspace_id, configuration.reimbursable_expenses_object))
+        create_ap_payment(workspace_id=expense_group.workspace.id)
```
ashwin1111, line 850: same comment as above

apps/sage_intacct/tasks.py
```diff
@@ -844,6 +846,8 @@ def create_bill(expense_group: ExpenseGroup, task_log_id: int, last_export: bool
         logger.exception('Something unexpected happened workspace_id: %s %s', task_log.workspace_id, task_log.detail)
     
     if last_export:
+        logger.info("| Last Export for Bill Creation | Content: {{WORKSPACE_ID : {} Configuration Expense Type: {}}}".format(expense_group.workspace_id, configuration.reimbursable_expenses_object))
```
ashwin1111, line 849: can remove both the loggers after local testing, doesn't add a lot of value since we log the payloads anyways

ashwin1111 (approved):

PR #446, opened 2024-03-05
Error Parser Model Changes, Core Logic

ruuushhh:


apps/sage_intacct/tasks.py
```diff
@@ -302,14 +306,28 @@ def handle_sage_intacct_errors(exception, expense_group: ExpenseGroup, task_log:
...
     Error.objects.update_or_create(
             workspace_id=expense_group.workspace_id,
             expense_group=expense_group,
             defaults={
                 'type': 'INTACCT_ERROR',
                 'error_title': error_title,
                 'error_detail': error_msg,
-                'is_resolved': False
+                'is_resolved': False,
+                'is_parsed': is_parsed,
```
Shwetabhk, line 328: is_parsed, attribute_type, and article link are initialized inside the if block at line 311 but is being accesses outside the scope of it, it will fail if the code doesn't not enter the if block, in general not a good practice
needs fixing

Shwetabhk (approved):

PR #472, opened 2024-03-26
Added logic to clean the task_logs, errors, last_export_detail on export type change

Hrishabh17:
## Summary by CodeRabbit

- **New Features**
	- Introduced functionality to automatically clean up task logs and errors for unselected exports in the workspace settings.
	- Added triggers for export settings to handle specific configurations for different expense groups within a workspace.
	- Implemented test cases to verify the behavior of post-save configuration triggers related to exporting settings.


apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -154,6 +154,15 @@ def update(self, instance, validated):
             }
         )
 
+        export_trigger = ExportSettingsTrigger(
+            workspace_id=workspace_id,
+            configuration=Configuration.objects.filter(
+                workspace_id=workspace_id
+            ).first()
```
ashwin1111, line 161: need not query db again, store 146 to a variable and use it

PR #726, opened 2025-06-27
fix: bugfixes in cursor-generated code

JustARatherRidiculouslyLongUsername:
## Clickup
app.clickup.com

## Summary by CodeRabbit

* **Refactor**
  * Simplified the E2E setup process to require only a workspace ID, removing the need for user and organization details.
  * Enhanced idempotency and reliability when updating or creating related settings and records.
  * Streamlined fixture creation and improved associations between test objects.
  * Improved error logging for setup and cleanup operations to include full exception tracebacks.


apps/internal/services/e2e_setup.py
```diff
@@ -45,68 +41,35 @@ def _setup_phase1_core_data(self):
...
-        )
-
-        # 2. Create auth_tokens
-        AuthToken.objects.create(
-            user=user,
-            refresh_token=self.refresh_token
-        )
+        # 1. Get workspace and rename it, marking it ready to delete
+        workspace = Workspace.objects.get(id=self.workspace_id)
+        workspace.name = 'E2E Integration Tests'
```
ashwin1111, line 46: on each login to integration we update the workspace name via background worker to keep it in-line with Fyle's org name, we may have to introduce other way to validate workspace name to reset, this I feel may not work as we expect

apps/internal/services/fixture_factory.py
```diff
@@ -2,32 +2,44 @@
...
     """Factory for creating Intacct test fixture data"""
 
-    def create_dependent_field_settings(self, workspace, count=2):
+    def create_dependent_field_settings(self, workspace):
         """Create sample dependent field settings"""
         settings = []
 
-        for i in range(count):
+        # Disconnect the pre_save signal to avoid triggering dependent field creation
+        pre_save.disconnect(run_pre_save_dependent_field_settings_triggers, sender=DependentFieldSetting)
```
ashwin1111, line 24: does this disconnect things only for this particular instance or the whole API server?

JustARatherRidiculouslyLongUsername: it did for the whole API - fixing everything in #735

PR #728, opened 2025-06-30
feat: Create vendor when 'Another record with the' value error

ruuushhh:
### Description
feat: Create vendor when 'Another record with the' value error

## Clickup
https://app.clickup.com/t/86czf7g0n

### DIFF
--- a/apps/sage_intacct/utils.py
+++ b/apps/sage_intacct/utils.py
@@ -1,52 +1,40 @@
-import re
 import json
 import logging
+import re
 from datetime import datetime, timedelta
 from typing import Optional
 
-from apps.workspaces.helpers import get_app_name
 import unidecode
+from cryptography.fernet import Fernet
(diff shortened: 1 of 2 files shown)


apps/sage_intacct/utils.py
```diff
@@ -1016,7 +1004,12 @@ def get_or_create_vendor(self, vendor_name: str, email: str = None, create: bool
 
                         vendor = vendor if vendor['STATUS'] == 'active' else None
                     else:
-                        vendor = None
+                        try:
+                            vendor_id = vendor_id + '-1'
+                            vendor = self.post_vendor(vendor_id, vendor_name, email)
+                        except Exception as e:
+                            logger.info("Error while creating vendor %s in Workspace %s: %s", vendor_name, self.workspace_id, e.response)
```
ashwin1111, line 1011: should be logger.error()

PR #757, opened 2025-07-24
feat: support using real creds in e2e setup

JustARatherRidiculouslyLongUsername:
## Clickup
app.clickup.com

### DIFF
--- a/apps/internal/views.py
+++ b/apps/internal/views.py
@@ -108,7 +108,8 @@ class E2ESetupView(generics.GenericAPIView):
         try:
             # Initialize setup service
             setup_service = E2ESetupService(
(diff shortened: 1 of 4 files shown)


apps/internal/views.py
```diff
@@ -106,9 +106,12 @@ def post(self, request: Request) -> Response:
         validated_data = serializer.validated_data
 
         try:
+            print(f'{validated_data=}')
+            print(f'{validated_data["use_real_intacct_credentials"]=}')
```
ashwin1111, line 110: remove loggers pls / make it logger.info()

JustARatherRidiculouslyLongUsername: done, this was temporary

